PR #35, opened 2018-06-04
Account for density in divergence check

bartvstratum:
Density was not accounted for in the divergence check; it should be `d(rho*u_i)/dx_i` instead of  `d(u_i)/dx_i`


julietbravo:
The first commit checked the continuity equations d(rho*u_i)/dx_i; as mentioned by Huug (e-mail), this resulted in the wrong units (not in s^-1). Second commit fixes this.

fjansson:
Thanks! This has been fixed, somehow I didn't close the pull request then. Closing now.

PR #39, opened 2018-09-25
include kind specifier in ceiling function

ipelupessy:
timeleft is longint, however ceiling (at least on gfortran 8.1.1) defaults to int32 leading to an incorrect value
(due to overflow) of timeleft.

fix is to include kind specifier in ceiling

### DIFF
--- a/src/modglobal.f90
+++ b/src/modglobal.f90
@@ -456,7 +456,7 @@ contains
       end do
     end if
     tnextrestart = trestart/tres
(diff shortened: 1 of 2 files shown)


ipelupessy:
added the same fix for itrestart
(not tested yet)

fjansson:
Done, in commit 39f75a78e. (This pull request was not merged due to other changes)

PR #45, opened 2019-07-31
Halo exchange (excjs) changes

jiskattema:
Send all requests asynchronously, and wait for them later.
This is possibly 4x faster then previous send/wait approach.

The code runs and compiles, but needs bug testing and performance testing.


jiskattema:
Thanks for testing! 
The latest commit forces the east-west exchange to wait for the north-south exchanges; this should take care of the issue you found.
As the amount of work happening async is now reduced, the speedup is now a factor of ~2 instead of ~4..

fjansson:
This version gives bitwise identical results as the old code on the bomex test above with 2x2 processes.

fjansson:
I got a segfault on Lisa, testing the code on a bigger case: in modpois.f90, the pup, pvp, pwp fields have only one halo cell, and used to be exchanged with excj. Now the new excjs call uses ih, jh halo cells, which results in reading and writing outside bounds. With bomex this passed undetected since there ih, jh = 1. Fix: replace ih, jh by 1 here.
```
    call excjs(pup,2,i1,2,j1,1,k1,1,1)
    call excjs(pvp,2,i1,2,j1,1,k1,1,1)
    ! call excjs(pwp,2,i1,2,j1,1,k1,1,1)
```

Even with this fix, the old and new halo exchange do not give bit-wise exact results on the larger test, velocities after 2h differ by ~ 1e-8. The test had 5x5 processes, over 2 Lisa nodes.
Neither the old nor the new code gave bitwise identical results on consecutive runs, the size of the difference between subsequent runs of the same code was of similar size as between the two versions. One possible reason is that the MPI reduction operations aren't reproducible when the job spans several nodes.
The new code gave a 6% decrease in wall-clock time for this test.

A small optimization: in excjs, we can add a check `if (ih!=0)` around exchanging in x, and `if (jh!=0)` around exchanging in y, it's easy to do the pup, pvp exchanges only in the needed direction.

fjansson:
I cannot get the optimization to work though. Adding the check seems OK, but omitting p{u,v}p exchange in one of the directions seems to break something and gives a thermo crash.

jiskattema:
Your optimization of ih=0 to disable east/west exchange fails because the value of ih is also used in the definition of the array shape in excjs. You would need to add another function argument to if you want to select ns, ew, or both.

fjansson:
Asynchronous send,receive was implemented some time back and seems beneficial. 
It was done in two stages to avoid corner issues: east and west in parallel, and north and south in parallel.
Closing this.

PR #55, opened 2020-06-29
modmpi: restore excj to build rico test case

dguibert:


fjansson:
@jiskattema can you see if the moduser of the rico case can be adapted to the new halo exchange functions, instead of bringing back the old ones?

jiskattema:
It's been a while since i looked at this, but i thought you can use excjs instead of excj.

You have to be a bit careful with the array sizes: the code started out with a hardcoded boundary size of 1, and together with Fortran arrays starting at 1, the data for arrays start at index 2.
Then there are imax points, so data ends at index imax+1 
=> for the first dimension we end up with [2, i1] as range

Then we started using Fortrans custom array bounds to implement (runtime dependent) halo zone: 
=> range for the first dimension [2-ih:i1+ih]

Applying this to the moduser call:
call excj( ustar  , 1, i2, 1, j2, 1,1)

where:
* allocate(ustar   (i2,j2))
* i2 and j2 are imax+2 and jmax+2

So the field ustar has a halo of size ih=jh=1, and is equivalent to ustar(2-1:i1+1, 2-1:j1+1)

For the excjs function you would get:
call excjs(ustar, 2, i1, 2, j1, 1, 1, 1, 1)

fjansson:
Fixed in 694bb28daf06, now in `main`.

PR #57, opened 2020-07-15
proposing a revised Deardorff SGS TKE scheme

zrxdaly:
Hi I would like to add a revise Deardorff SGS TKE scheme to the repository ([publication](https://arxiv.org/abs/2003.09463) is  provisionally accepted with minor revisions in Boundary Layer Meteorology).

The main revision is from  $$\lambda = min(\Delta, L_b)$$ to $$\frac{1}{\lambda}=\frac{1}{\kappa z}+\frac{1}{L_{b}}$$ 

There are changes in three files:
modbudget.f90: changes to add output of mixing length scale 

modsubgrid.f90: changes to parameterization of $$\lambda$$

modtimedep.f90: changes to fix a small bugs which is already addressed in branch to4.3_Fredrik

### DIFF
--- a/src/modbudget.f90
+++ b/src/modbudget.f90
@@ -33,7 +33,7 @@ module modbudget
   PUBLIC :: initbudget,budgetstat,exitbudget
   save
 !NetCDF variables
-  integer,parameter :: nvar = 18
+  integer,parameter :: nvar = 19
   character(80),dimension(nvar,4) :: ncname
 
   real    :: dtav, timeav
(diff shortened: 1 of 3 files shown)


huugouwersloot:
Is this really a bug, or rather a (scientific) improvement?
Especially if it is the latter case, but preferrably also in the first, modify the code changes such that a switch is required to toggle between the original and the new representation, with default value to have the original representation.

zrxdaly:
> Is this really a bug, or rather a (scientific) improvement?
> Especially if it is the latter case, but preferrably also in the first, modify the code changes such that a switch is required to toggle between the original and the new representation, with default value to have the original representation.

It is an improvement on Deardorff SGS TKE scheme for the simulation on Stable boundary layer. More details and differences in representation between the default scheme and revised scheme can be seen in the [paper](https://arxiv.org/abs/2003.09463).

huugouwersloot:
In that case, please rewrite the code to make application of the improvement optional

PR #60, opened 2020-10-05
Feature/uvw m0p 4

v1kko:
A draft for merging 4bytes floating points into DALES.

If we decide to do this, then there is still some work needed to make all of this optional under a preprocessor flag


v1kko:
@jiskattema

PR #64, opened 2021-03-17
Feature/explicit interface

v1kko:
- Creating explicit interfaces allows compilation without ``std=legacy`` and with ``-Werror=implicit-interface -std=gnu``
- Implicit conversion errors between subroutines are caught at compile time

### DIFF
--- a/src/fftnew.f90
+++ b/src/fftnew.f90
@@ -17,6 +17,8 @@
 !
 !  Copyright 1993-2009 Delft University of Technology, Wageningen University, Utrecht University, KNMI
 !
+module fftnew
+contains
 
   subroutine RADB2 (IDO,L1,CC,CH,WA1)
       DIMENSION       CC(IDO,2,L1)           ,CH(IDO,L1,2), WA1(IDO-1)
@@ -728,6 +730,7 @@
   end
 
   subroutine RFFTB1 (N,C,CH,WA,IFAC)
+    REAL ::  IFAC
       DIMENSION       CH(N)      ,C(N)       ,WA(N+1)      ,IFAC(2*N+1)
   NF = IFAC(2)
   NA = 0
@@ -806,6 +809,7 @@
   end
 
   subroutine RFFTF1 (N,C,CH,WA,IFAC)
+    real :: IFAC
       DIMENSION       CH(N)      ,C(N)       ,WA(N+1)      ,IFAC(2*N+1)
   NF = IFAC(2)
   NA = 1
@@ -883,6 +887,7 @@
   end
 
   subroutine RFFTI1 (N,WA,IFAC)
+  real :: IFAC
   logical firstloop
   DIMENSION       WA(N+1)      ,IFAC(2*N+1)    ,NTRYH(4)
   DATA NTRYH(1),NTRYH(2),NTRYH(3),NTRYH(4)/4,2,3,5/
@@ -955,3 +960,4 @@
   end do
   return
   end
+end module fftnew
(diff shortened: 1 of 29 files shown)


src/fftnew.f90
```diff
@@ -728,6 +730,7 @@ subroutine RFFTB (N,R,WSAVE)
   end
 
   subroutine RFFTB1 (N,C,CH,WA,IFAC)
+    REAL ::  IFAC
```
v1kko, line 733: This is the only change in this pull request I am not confident in, it seems to work, but maybe we have to make the conversion from real to integer explicit

v1kko:
@fjansson What do you think?

src/fftnew.f90
jiskattema: I'm not sure what the original code is doing; as i understand it there are two options:
1. as everything in fortran is passed by reference, the cast is from an int pointer to a real pointer. the actual value is never cast, and always has full precision. Writing as int and reading as real will give very weird results, but i guess they dont do that.
2. there is some implicit casts going on, and your suggested change makes them explicit.

A way to test whjich one it is, is to write out the values of ifac / wsave before and after the function calls.

fjansson:
Without looking in detail, I remember discovering that the code uses some array, maybe the work / temp array, to store both reals and ints. I think that if your change would be wrong, it would break the results in an obvious way. I like the code being more explicit.

This pull request is just the interfaces and github actions, right, i.e. no single precision yet? It seems it should be safe to merge. I'll try to do it, tomorrow or early next week.

src/fftnew.f90
v1kko: Hi Jisk, it is interesting and I think 1. is what happens. However no place is both used as an integer and rea (I think). I have created a new commit which makes all interfaces in fftnew explicit, but which explicitly uses a wrong interface here (which also happens implicitly)

src/fftnew.f90
```diff
@@ -17,8 +17,69 @@
...
+  subroutine RFFTB (N,R,WSAVE)
+    interface 
+      subroutine RFFTB1 (N, C, CH, WA, WFAC)
+        use iso_fortran_env, only : real64, int64
+        implicit real(real64)   (a-h)
+        implicit real(real64)   (o-z)
+        implicit integer(int64) (i-n)
+      DIMENSION       CH(N)      ,C(N)
+      end subroutine
+    end interface
```
v1kko, line 37: These changes should make all implicit rules explicit, while retaining the implicit cast @jiskattema

src/fftnew.f90
```diff
@@ -50,6 +111,10 @@ subroutine RADB2 (IDO,L1,CC,CH,WA1)
   end
 
   subroutine RADB3 (IDO,L1,CC,CH,WA1,WA2)
+      use iso_fortran_env, only : real64, int64
+      implicit real(real64)   (a-h)
+      implicit real(real64)   (o-z)
+      implicit integer(int64) (i-n)
```
v1kko, line 117: I have added the implicit rules everywhere to at least have same-sized integers and reals

src/fftnew.f90
```diff
@@ -720,14 +817,50 @@ subroutine RADFG (IDO,IP,L1,IDL1,CC,C1,C2,CH,CH2,WA)
...
-  return
-  end
+
 
   subroutine RFFTB1 (N,C,CH,WA,IFAC)
+      use iso_fortran_env, only : real64, int64
+      implicit real(real64)   (a-h)
+      implicit real(real64)   (o-z)
+      implicit integer(int64) (i-n)
+      interface
```
v1kko, line 827: Since these subroutines are outside of a module, they need interfaces to all other routines that are called

PR #66, opened 2021-03-24
Feature/explicit hypre

v1kko:
This pull request makes everything in hypre explicit

- one bug is revealed (the only error left), How should this be fixed? @fjansson , @jiskattema 

- Perhaps we can link it to the HYPRE headers to always correctly use singles/doubles/complex etc if HYPRE is not compiled with default values


fjansson:
```
/home/runner/work/dales/dales/src/modhypre.f90:1044:63:

 1044 |       call HYPRE_StructPCGSetRelChange(solver, tolerance, ierr)
      |                                                               1
Error: Type mismatch in argument ‘rel_change’ at (1); passed REAL(8) to INTEGER(4)
```

https://github.com/ESiWACE-S1/dales/blob/f51336768da3543761ae4ffbdcd02d7596ea392b/src/modhypre.f90#L1044

This looks like a bug by us - we pass a real tolerance value to a function which expects an int. I think the parameter to the function isn't a tolerance but a flag - this is from the HYPRE manual for a different solver with a similarly named function:
```
StructPFMGSetRelChange:
(Optional) Additionally require that the relative difference in successive iterates be small.
```

I think the solution is to instead call 
```
HYPRE_Int HYPRE_StructPCGSetTol(HYPRE_StructSolver solver,  HYPRE_Real tol)
```

jiskattema:
> ```
> /home/runner/work/dales/dales/src/modhypre.f90:1044:63:
> 
>  1044 |       call HYPRE_StructPCGSetRelChange(solver, tolerance, ierr)
>       |                                                               1
> Error: Type mismatch in argument ‘rel_change’ at (1); passed REAL(8) to INTEGER(4)
> ```
> 
> https://github.com/ESiWACE-S1/dales/blob/f51336768da3543761ae4ffbdcd02d7596ea392b/src/modhypre.f90#L1044
> 
> This looks like a bug by us - we pass a real tolerance value to a function which expects an int. I think the parameter to the function isn't a tolerance but a flag - this is from the HYPRE manual for a different solver with a similarly named function:
> 
> ```
> StructPFMGSetRelChange:
> (Optional) Additionally require that the relative difference in successive iterates be small.
> ```
> 
> I think the solution is to instead call
> 
> ```
> HYPRE_Int HYPRE_StructPCGSetTol(HYPRE_StructSolver solver,  HYPRE_Real tol)
> ```

Yes, that is the obvious fix!

v1kko:
I fixed the bug, and I think it can be merged

fjansson:
After this change in CMakeLists.txt 
```
-FIND_LIBRARY(HYPRE_LIB HYPRE
+FIND_LIBRARY(HYPRE_LIB HYPRE_core)
```
the automatic finding of HYPRE doesn't work for me any more on Arch Linux. I think it's possible to specify several possible names for the library, I will try that.

Any idea why `use mpi,             only : mpi_allreduce` breaks the compile with Intel?

v1kko:
@fjansson Maybe it is good to specify both HYPRE and HYPRE_core, HYPRE_core can then be an optional library
- in ubuntu the HYPRE lib is present, but doesn't include the fortran bindings, however linking to both is not a problem

I think that mpi line breaks because `use mpi` and `include 'mpif.h` are very old and probably do not specify the interfaces. My next pull request will fix this by migrating everything to mpi_f08 :)

fjansson:
Ok, good to know. Listing both library names works for me. I think `HYPRE_core` should be listed first, so that it is picked if both exist like on Ubuntu. 

mpi_f08 sounds great too!

PR #67, opened 2021-04-12
Feature/d mpi

v1kko:
- This pull request upgrades all MPI to mpi_f08 such that explicit interfaces can be used

- All MPI calls are transformed in D_MPI calls.
  The D_MPI interface is  a "Dales" wrapper around MPI and makes sure that the correct number of bytes is sent and received.

- This pull request changes a lot, so it might be good to run some tests to validate it


v1kko:
@fjansson  The next pull request :)

src/modfftw.f90
```diff
@@ -61,6 +61,15 @@ module modfftw
...
 contains
+  subroutine static_checks_dontcall
+    use iso_fortran_env, only : real64
+    implicit none
+    real                :: check
+! Statically check for correct double precisions, FFTW C-library breaks without it
+    check = sign(1._real64, p201(1,1,1))
+    check = sign(1._real64, p210(1,1,1))
+    check = sign(1._real64, p_nohalo(1))
+  end subroutine
```
v1kko, line 72: This should be removed

v1kko: removed and force pushed

fjansson:
Does the commit "excjs use non-blocking receive and subslicing" allow north-south exchanges to happen in parallel with east-west exchanges? 

I think that would break the exchange of corner elements. Some corner elements are required (because of how the velocity grids are defined). North and south can be done in parallel. East and west can also be done in parallel. Jisk and I tried once to do all the exchanges in parallel, and then we discovered the corner issue. In principle we could do corners separately (perhaps for just the fields that need them), but it feels complicated.

I like the reshape construction.

v1kko:
@fjansson you are right that I was sloppy in the excjs changes with relation to the corner, I rewrote them such that it should be clear that the corners are communicated correctly, but please check (especially that change) thoroughly

fjansson:
Thanks, I'll check and test. One idea: for the fields where the corners aren't needed, it might be faster to do all four directions in parallel. Right now I think only u,v and maybe w need corner elements, this should be checked of course.

v1kko:
@fjansson I found a bug which caused the real64 to be communicated as real32, I am currently testing the version with the latest bugfix on cartesius

fjansson:
The Fujitsu compiler doesn't accept 58093c1bba. gfortran 8 and 10, and ifort on Cartesius compiled it happily, though.

```
Fortran diagnostic messages: program name(modmpi)
  jwd2278i-w  "/home/hp120279/u00892/dales-tester/dales/src/modmpi.f90", line 71: Specific procedures (D_MPI_ISEND_REAL32) and (D_MPI_ISEND_REAL64) do not ensure that generic reference (D_MPI_ISEND) is unambiguous.
  jwd2278i-w  "/home/hp120279/u00892/dales-tester/dales/src/modmpi.f90", line 75: Specific procedures (D_MPI_IRECV_REAL32) and (D_MPI_IRECV_REAL64) do not ensure that generic reference (D_MPI_IRECV) is unambiguous.
  jwd2278i-w  "/home/hp120279/u00892/dales-tester/dales/src/modmpi.f90", line 79: Specific procedures (D_MPI_BCAST_REAL32) and (D_MPI_BCAST_REAL64) do not ensure that generic reference (D_MPI_BCAST) is unambiguous.
  jwd2278i-w  "/home/hp120279/u00892/dales-tester/dales/src/modmpi.f90", line 86: Specific procedures (D_MPI_ALLREDUCE_REAL32) and (D_MPI_ALLREDUCE_REAL64) do not ensure that generic reference (D_MPI_ALLREDUCE) is unambiguous.
  jwd2278i-w  "/home/hp120279/u00892/dales-tester/dales/src/modmpi.f90", line 86: Specific procedures (D_MPI_ALLREDUCE_REAL32_S) and (D_MPI_ALLREDUCE_REAL64_S) do not ensure that generic reference (D_MPI_ALLREDUCE) is unambiguous.
  jwd2278i-w  "/home/hp120279/u00892/dales-tester/dales/src/modmpi.f90", line 93: Specific procedures (D_MPI_ALLTOALL_REAL32) and (D_MPI_ALLTOALL_REAL64) do not ensure that generic reference (D_MPI_ALLTOALL) is unambiguous.
  jwd2278i-w  "/home/hp120279/u00892/dales-tester/dales/src/modmpi.f90", line 97: Specific procedures (D_MPI_REDUCE_REAL32) and (D_MPI_REDUCE_REAL64) do not ensure that generic reference (D_MPI_REDUCE) is unambiguous.
  jwd2278i-w  "/home/hp120279/u00892/dales-tester/dales/src/modmpi.f90", line 97: Specific procedures (D_MPI_REDUCE_REAL32_S) and (D_MPI_REDUCE_REAL64_S) do not ensure that generic reference (D_MPI_REDUCE) is unambiguous.
  jwd2278i-w  "/home/hp120279/u00892/dales-tester/dales/src/modmpi.f90", line 103: Specific procedures (D_MPI_GATHER_REAL32) and (D_MPI_GATHER_REAL64) do not ensure that generic reference (D_MPI_GATHER) is unambiguous.
  jwd2278i-w  "/home/hp120279/u00892/dales-tester/dales/src/modmpi.f90", line 108: Specific procedures (excjs_real32) and (excjs_real64) do not ensure that generic reference (excjs) is unambiguous.
  jwd2278i-w  "/home/hp120279/u00892/dales-tester/dales/src/modmpi.f90", line 112: Specific procedures (slabsum_real32) and (slabsum_real64) do not ensure that generic reference (slabsum) is unambiguous.
 Module subprogram name(excjs_real32)
  jwd2520i-s  "/home/hp120279/u00892/dales-tester/dales/src/modmpi.f90", line 280, column 10: Generic procedure reference must be unambiguous for 'D_MPI_ISEND'.
  jwd2520i-s  "/home/hp120279/u00892/dales-tester/dales/src/modmpi.f90", line 281, column 10: Generic procedure reference must be unambiguous for 'D_MPI_ISEND'.
  jwd2520i-s  "/home/hp120279/u00892/dales-tester/dales/src/modmpi.f90", line 284, column 10: Generic procedure reference must be unambiguous for 'D_MPI_IRECV'.
  jwd2520i-s  "/home/hp120279/u00892/dales-tester/dales/src/modmpi.f90", line 285, column 10: Generic procedure reference must be unambiguous for 'D_MPI_IRECV'.
  jwd2520i-s  "/home/hp120279/u00892/dales-tester/dales/src/modmpi.f90", line 312, column 10: Generic procedure reference must be unambiguous for 'D_MPI_ISEND'.
  jwd2520i-s  "/home/hp120279/u00892/dales-tester/dales/src/modmpi.f90", line 313, column 10: Generic procedure reference must be unambiguous for 'D_MPI_ISEND'.
  jwd2520i-s  "/home/hp120279/u00892/dales-tester/dales/src/modmpi.f90", line 316, column 10: Generic procedure reference must be unambiguous for 'D_MPI_IRECV'.
  jwd2520i-s  "/home/hp120279/u00892/dales-tester/dales/src/modmpi.f90", line 317, column 10: Generic procedure reference must be unambiguous for 'D_MPI_IRECV'.
 Module subprogram name(excjs_real64)
  jwd2520i-s  "/home/hp120279/u00892/dales-tester/dales/src/modmpi.f90", line 394, column 10: Generic procedure reference must be unambiguous for 'D_MPI_ISEND'.
  jwd2520i-s  "/home/hp120279/u00892/dales-tester/dales/src/modmpi.f90", line 395, column 10: Generic procedure reference must be unambiguous for 'D_MPI_ISEND'.
  jwd2520i-s  "/home/hp120279/u00892/dales-tester/dales/src/modmpi.f90", line 398, column 10: Generic procedure reference must be unambiguous for 'D_MPI_IRECV'.
  jwd2520i-s  "/home/hp120279/u00892/dales-tester/dales/src/modmpi.f90", line 399, column 10: Generic procedure reference must be unambiguous for 'D_MPI_IRECV'.
  jwd2520i-s  "/home/hp120279/u00892/dales-tester/dales/src/modmpi.f90", line 426, column 10: Generic procedure reference must be unambiguous for 'D_MPI_ISEND'.
  jwd2520i-s  "/home/hp120279/u00892/dales-tester/dales/src/modmpi.f90", line 427, column 10: Generic procedure reference must be unambiguous for 'D_MPI_ISEND'.
  jwd2520i-s  "/home/hp120279/u00892/dales-tester/dales/src/modmpi.f90", line 430, column 10: Generic procedure reference must be unambiguous for 'D_MPI_IRECV'.
  jwd2520i-s  "/home/hp120279/u00892/dales-tester/dales/src/modmpi.f90", line 431, column 10: Generic procedure reference must be unambiguous for 'D_MPI_IRECV'.
 Module subprogram name(gatherrow)
  jwd2520i-s  "/home/hp120279/u00892/dales-tester/dales/src/modmpi.f90", line 557, column 10: Generic procedure reference must be unambiguous for 'D_MPI_GATHER'.
```

v1kko:
Hmm, perhaps with our combination of compile flags, real32 == real64, then the two routines would be ambiguous.

Can tests the following simple program with the dales compile flags?:
`program main
use iso_fortran_env
real(real32) a
real(real64) b
write(*,*) sizeof(a), sizeof(b)
end program`

fjansson:
The test program outputs `8 8` when compiled with the Fujitsu compiler and with the flags we use for DALES, in particular `-Ad`  (sets the default real precision to 64 bit). Compiled with no flags, the program prints  `4 8`.

v1kko:
I think the  -CcdRR8 needs to be used instead of the -Ad flag

fjansson:
With `-CcdRR8` your test program prints `4 8` so that looks promising. I'll try it on DALES too.

fjansson:
DALES compiles with -CcdRR8. I pushed this change to to4.4_Fredrik, 115bcb5655c.

Now I'm seeing warnings about ALOG10, ALOG, AMIN1, AMAX1, e.g.
```
 jwd2207i-w  " xxxxx/dales/src/modsimpleice.f90", line 193, column 24: Invalid type of actual argument for intrinsic function 'AMAX1'.
```
gnu fortran manual says AMAX1 : "Archaic form of MAX() that is specific to one type for A". I guess we can replace all of these with log10, log, min, max.

jiskattema:
> gnu fortran manual says AMAX1 : "Archaic form of MAX() that is specific to one type for A". I guess we can replace all of these with log10, log, min, max.

Some googling gives the same opinion: the explicit one should no longer be used.
So yes replace ```AMAX1``` with ```MAX```

fjansson:
I replaced all amin, amax etc, now in to4.4_Fredrik. 

One more warning: 
`jwd2207i-w  "/home/hp120279/u00892/dales-tester/dales/src/fftnew.f90", line 652, column 13: Invalid type of actual argument for intrinsic function 'FLOAT'.`
This is about  `ARG = TPI/FLOAT(IP)`
TPI is real already a real (2*pi). IP is int64. Maybe we can just drop the float() ?

v1kko:
Yes I agree that we can drop the float and let the compiler handle the promotion to real.

fjansson:
I removed the float() functions, pushed to to4.4_Fredrik, testing now. Seems to compile without warnings at least.

I tried merging ESiWACE-S1:feature/d_mpi into to4.4_Fredrik, and compile with Fujitsu. Still not working, compiling with -CcdRR8 :
```
Fortran diagnostic messages: program name(modmpi)
 Module subprogram name(D_MPI_ISEND_REAL32)-CcdRR8
  jwd2516i-s  "/home/hp120279/u00892/dales/src/modmpi.f90", line 582, column 10: Reference to 'mpi_isend' not consistent with any specific interface of the generic interface.
 Module subprogram name(D_MPI_IRECV_REAL32)
  jwd2516i-s  "/home/hp120279/u00892/dales/src/modmpi.f90", line 599, column 10: Reference to 'mpi_irecv' not consistent with any specific interface of the generic interface.
 Module subprogram name(D_MPI_BCAST_REAL32)
  jwd2516i-s  "/home/hp120279/u00892/dales/src/modmpi.f90", line 615, column 10: Reference to 'mpi_bcast' not consistent with any specific interface of the generic interface.

...
```
Maybe buf of type real32 doesn't match any of the MPI functions. I don't see any mention of the real32 type in the Fujitsu MPI manuals.

v1kko:
I think real32 gets interpreted as '4'. but you can try if it compiles with '4' instead of 'real32'.

Is the mpi library source available?
Otherwise my guess is that in the mpi_isend irecv and bcast routines in the mpi_f08 module, no kind is specified for the buf variable. We could try to fix that by declaring it external, another option might be to use the old mpi interface

v1kko:
perhaps rewriting the routine as something like:
```fortran
  subroutine D_MPI_ISEND_REAL32(buf, count, dest, tag, comm, request, ierror)
    implicit none
    interface
      subroutine MPI_ISEND(buf, count, datatype, dest, tag, comm, request, ierror)
        use mpi_f08, mpi_isend_faulty => mpi_isend
        TYPE(*), DIMENSION(..), INTENT(IN), ASYNCHRONOUS :: buf
        INTEGER, INTENT(IN) :: count, dest, tag
        TYPE(MPI_Datatype), INTENT(IN) :: datatype
        TYPE(MPI_Comm), INTENT(IN) :: comm
        TYPE(MPI_Request), INTENT(OUT) :: request
        INTEGER, OPTIONAL, INTENT(OUT) :: ierror
      end subroutine
    end interface
    real(real32), asynchronous ::   buf(..)
    integer       ::   count, dest, tag, ierror
    type(MPI_COMM):: comm
    type(MPI_REQUEST) :: request
    call MPI_ISEND(buf,count,MPI_REAL4,dest,tag,comm,request,ierror)
  end subroutine D_MPI_ISEND_REAL32
```

Although that might disable some optimizations within the mpi routine...

v1kko:
I guess the mpi_isend is implemented with an interface, with generics for integer(4), integer(8) etc. However I think they incorrectly used real instead of real(4). which with -CCdrr8 becomes real(8). We could raise an issue, the fix should be simple on their side

v1kko:
Then perhaps they forgot to implement real(4). Although it makes me wonder why they just didn't use type(*) instead of an interface. However it should be easy enough to fix in the library.

Perhaps it is implemented this way because they want to support MPI_SUBARRAYS_SUPPORTED, and MPI_ASYNC_PROTECTS_NONBLOCKING . Which, if supported, would mean that we do not need buffers for communication anymore, and would enable extra optimization possibilities :)

v1kko:
Hi @fjansson, I merged the fixes from your branch and rebased, I kept the pointer casting as the fusjitsu compiler does not define a macro through which it is identifiable (at least I couldn't find it). Instead, there is a comment in the code that the casting can be removed whenever the fujitsu compiler is fixed

fjansson:
e57b552 compiles on Fugaku, but seems to hang when running the 48h RICO test. It hangs in the very beginning, after printing 
```
 &SOLVER SOLVER_ID=100,MAXITER=10000,TOLERANCE=1.000000000000000E-08,N_PRE=1,N_POST=1,PRECOND=1/
MPI mesh nprocx, nprocy:  6 8
```
I submitted a test run on Cartesius for the same commit.

fjansson:
...and the Cartesius run finished, looking ok.

v1kko:
@fjansson This is now contained in the new pull request #73

PR #69, opened 2021-06-08
make sure tag_git_version is build

v1kko:


v1kko:
@fjansson, this is a small dependency problem I encountered on fugaku

fjansson:
We have this fix in commit 56d5f0add716511f86 on the master branch - apparently I never applied that in to4.4_Fredrik. I think it's cleaner if I merge master into to4.4_Fredrik instead of this PR. Another consequence of merging is that the dales binary is named dales4.3 instead of dales4

PR #83, opened 2023-10-18
Update cases/rico/moduser.f90 for changes to excjs.

esclapez:
The call to the halo exchange function in rico/moduser.f90 was not up to date with the interface version introduced to handle multiple precision.

### DIFF
--- a/cases/rico/moduser.f90
+++ b/cases/rico/moduser.f90
@@ -120,7 +120,7 @@ subroutine surf_user
 
   ! excjs is a 3D function, reshape ustar
   ustar_3D(1:i2, 1:j2, 1:1) => ustar
-  call excjs(ustar_3D, 2, i2, 2, j2, 1, 1, 1, 1)
+  call excjs(ustar_3D, 1, i2, 1, j2, 1, 1, 1, 1)
 
   do n=1,nsv
     do j=2,j1
--- a/cases/rico/moduser.f90
+++ b/cases/rico/moduser.f90
@@ -52,7 +52,7 @@ subroutine surf_user
  use modsurfdata,only : ustar,dudz,dvdz,dqtdz,dthldz,&
                           svs,z0,qts,thls,thvs,thlflux,qtflux,svflux
   use modfields, only : u0,v0,thl0,qt0,sv0,u0av,v0av,qt0av,thl0av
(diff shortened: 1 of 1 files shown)


fjansson:
Thanks! I will check this, and then merge into a development branch (probably to4.4.2_Fredrik)

The `rico` case under `cases/benchmarks/` uses one of the standard surface schemes, so with that you don't have to compile in the case-specific moduser.f90 file. It would great to get this fixed anyway. 

For OpenACC this moduser might need more changes, like some openacc tags?

esclapez:
Yes, it'll need more changes for OpenACC, but I didn't want to mix things up. I ran into this when I ran the `cases/rico` instead of the `cases/benchmark/rico`.

fjansson:
I don't understand the second commit. I think the exchange should be similar to the one in modsurface.f90?

This seems to work for me:
```
ustar_3D(1:i2, 1:j2, 1:1) => ustar
call excjs(ustar_3D, 2, i1, 2, j1, 1, 1, 1, 1)
```


Either way (first commit or both), I get a crash at 60 seconds when netCDF is trying to write somethin it says it cannot represent (probably nan or inf).

Also, these lines from moduser.f90 should be removed, I think they are from the time before MPI decomposition in both x and y, and ensure periodic boundaries.
```
 do j=2,j1
    ustar(1,j)=ustar(i1,j)
  end do
```

I'll have to test some more.

jchylik:
> Yes, it'll need more changes for OpenACC, but I didn't want to mix things up.

Is the moduser really needed in the rico case?

> , I think they are from the time before MPI decomposition 

It seems like it was for some significantly older version of dales. I have recently run an alternative version of rico that does not require any code modifications.

fjansson:
I've also been running RICO with the default surface scheme, in the cases/benchmark folder there is such a case. But the original RICO intercomparison specifies a certain simple surface scheme for the models to be as comparable as possible, and that's what is in the moduser.f90 file. The idea now was to finally bring it up to date, so that the example can run as intended. I actually don't know if it makes any difference for the results :)

fjansson:
I merged this here: 694bb28daf06b3 into branch to4.4.2_Fredrik, with the fixes I mentioned above.

PR #84, opened 2023-11-18
Fix Intel build in CI/CD pipeline

CasparJungbacker:
Intel released a new version of the oneAPI toolkit (version 2024) and this broke the builds in the Intel CI/CD pipelines. Fixed by removing the manual copies of `libimf`, `libintlc`, `libsvml` and `libirng`. 

I also changed the compiler for NetCDF from `ifort` to the newer `ifx`, since `ifort` [is set to be discontinued in 2024](https://www.intel.com/content/www/us/en/developer/articles/release-notes/intel-oneapi-hpc-toolkit-release-notes.html).

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -71,17 +71,6 @@ jobs:
           sudo ln -s `which ifx` /usr/bin/ifx
           cd netcdf-fortran
           NCDIR=/usr FC=/usr/bin/ifx ./configure
-          echo "Check path:"
-          ls /opt/intel/oneapi/compiler/latest/lib
-          #fix weird bug where libmf et al are not found
-          #sudo cp /opt/intel/oneapi/compiler/latest/lib/libimf.so \
-           #  /opt/intel/oneapi/compiler/latest/linux/lib/
-          #sudo cp /opt/intel/oneapi/compiler/latest/linux/compiler/lib/intel64_lin/libintlc.so.5 \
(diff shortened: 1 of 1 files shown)


CasparJungbacker:
@fjansson if you want to merge this, maybe squash the commits. They're not very neat 😅

fjansson:
Nice, thank you for addressing this! I saw the intel compiles started failing recently in the tests.
Are you sure ifx is enabled also when compiling DALES itself?
If I go here: 
https://github.com/CasparJungbacker/dales/actions/runs/6914838259/job/18813095193
and expand the "build" section, the warnings come from "ifort":
```
ifort: command line warning #10434: option '-extend_source' use with underscore is deprecated; use '-extend-source' instead
```
I guess we also need a CMake case for SYST=ifx to use ifx?
If you make one, maybe change `-extend_source` to `-extend-source` too. And let's keep the `ifort` option for a while for backwards compatibility (I mean only in CMake for compiling DALES, moving the automatic tests to ifx is fine by me).

CasparJungbacker:
Yes, DALES is still being compiled with `ifort` because `SYST=Cartesius` is used for CMake. According to [Intel's documentation](https://www.intel.com/content/www/us/en/docs/mpi-library/developer-reference-linux/2021-8/compilation-command-options.html), we would need to provide the flag `-fc=ifx` to `mpiifort` to switch to the `ifx` compiler. I'll try it when I have some time.

PR #90, opened 2024-01-09
Fix typo in bulkmicro accessing uninitialized qcbase in place of qrbase.

esclapez:


fjansson:
Good find, merged.

It seems on gfortran, the code worked, because qcbase kept it's value from the previous call, and qcbase (the lowest cloudy point) will generally be lower than the highest rainy point which is what should be stored in qrroof.

PR #98, opened 2024-05-27
OpenACC directives

esclapez:
- Move OpenACC porting into dev.


CasparJungbacker:
Hmmm, tests are failing in `modstartup.f90/readinitfiles` with message:

```Program received signal SIGFPE: Floating-point exception - erroneous arithmetic operation.```

@fjansson looks like the same NetCDF/HDF5 error we saw previously in debug builds?

fjansson:
It looks like the crash now is in icethermo and not from HDF5 this time. That crash would happen when opening a netcdf file.

CasparJungbacker:
I'm having the same issue on my local machine and Snellius. Release build works though.

esclapez:
Okay, got it. A couple of things missing from the refactoring of the surface schemes.

PR #101, opened 2024-06-06
4.4 ruisdael deposition

CasparJungbacker:
This merge introduces:

- Open boundary conditions
- Ruisdael deposition routines
- New land surface model
- Scalar/tracer manager

@fjansson anything else I forgot?


CasparJungbacker:
Merging `modsurface` is a mess, because of the clean-up on `dev`. Needs some testing, especially the with option `ltskininp=.true.`

fjansson:
@CasparJungbacker  for modsurface, there were only small changes introduced in the Ruisdael branch. 
I'd say keep the version from dev and add the other stuff by hand.

 446679fc57e97 
- some calls for the new LSM if that's selected
- add some pure keywords on functions

0a2259f85a8d5b
- one open boundaries exchange

01e9defe1
- initialize with patterned surface temperature

CasparJungbacker:
~Slight problem: looks like the new scalar manager is not backwards compatible. I.e., you _have_ to use a `tracerdata.inp` file for any run with scalars, even rico.~ 

You _can_ run without `tracerdata.inp`, but you have to provide some options in the namelist:

```
&NAMTRACERS
ltracers = .true.
tracernames = 'Nr', 'qr'
/
```

Bomex seems to run nicely though :-)

fjansson:
I'm fine with having to add a tracers section to the namoptions file for cases that use the scalars. Then we should just do it for all the sample cases that need it, and mention it in the documentation and release notes. Or did you make it do even more by default if tracers are not specified?

CasparJungbacker:
> I'm fine with having to add a tracers section to the namoptions file for cases that use the scalars. Then we should just do it for all the sample cases that need it, and mention it in the documentation and release notes. Or did you make it do even more by default if tracers are not specified?

Yeah now that I think of it, that might be better. In my latest commit, scalars are named `svx` if `tracernames` is left empty. That works for the benchmark rico case, because the columns in `scalar.inp` are called `sv0` and `sv1`. If they'd have different names, everything crashes again.

CasparJungbacker:
Time series look alright, ~although we do seem to have a bug in `vtke`~ (already fixed). Also the surface temperature flux (`wtheta`) seems to drift away from `4.4.2` after about 35 hours.

rico

fjansson:
the wtheta difference looks a little worrying. Could you add dev to the comparison and perhaps more ensemble members?

CasparJungbacker:
Now 10 members per ensemble. `wtheta` has definitely changed. 

`vtke` is also still wrong, but I think that's due to some missing parenthesis.

image

CasparJungbacker:
Ok, everything looks good now.

image

PR #105, opened 2024-07-23
Dev micro limiter fix

fjansson:
Change handling of negative qr - don't attempt to correct it with qt and thl adjustments.
Also correct plw, pli stats in genstat (used the wrong field).


fjansson:
there's probably an error still in modgenstat, some statistics (at least pli, plw, cli, clw) are not calculated for the lowest layer.
https://github.com/dalesteam/dales/blob/d9b261d9f34c8008dd0196de6ef26eb614a2ef4d/src/modgenstat.f90#L707

CasparJungbacker:
LGTM 👍

CasparJungbacker (approved):

PR #107, opened 2024-08-08
Fixes for OpenACC

esclapez:
- a fix in the genstat that was leading to occasional FP error on GPU
 - a fix in the warm restart on GPU that was leading to loosing the MPI rank <-> GPU pairing.


CasparJungbacker (approved):
Cool! That modgenstat issue was really annoying, couldn't find it myself 😅

PR #108, opened 2024-08-12
Fixes for running with open boundary conditions over land

fjansson:
*land surface model*: fix sign of rain flux, add some limits and diagnostics.
*microphysics*: run only up to kmax. If microphysics is done at k1, and open boundaries keep the k1 layer raining, it can form an "infinite" rain source, since cloud water is not removed from k1.
*modgenstat*: don't use `halflev_kappa` with open boundaries, because it can cause out-of bounds accesses at the top when w != 0. This change only affects statistics, not the time stepping.

### DIFF
--- a/src/tstep.f90
+++ b/src/tstep.f90
@@ -180,7 +180,7 @@ subroutine tstep_update
   !$acc parallel loop collapse(3) default(present) async(1)
   do k = 1, k1
     do j = 2, j2     ! i2, j2 here to include one ghost cell,
-      do i = 1, i2   ! needed for up, vp with open boundaries
+      do i = 2, i2   ! needed for up, vp with open boundaries
         up(i,j,k)=0.
         vp(i,j,k)=0.
         wp(i,j,k)=0.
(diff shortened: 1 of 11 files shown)


src/tstep.f90
```diff
@@ -179,8 +179,8 @@ subroutine tstep_update
   ! set all tendencies to zero
   !$acc parallel loop collapse(3) default(present) async(1)
   do k = 1, k1
-    do j = 2, j1
-      do i = 1, i1
+    do j = 2, j2     ! i2, j2 here to include one ghost cell,
```
esclapez, line 182: Throughout both `tstep_****` routines, `j` starts at 2 while `i` starts at 1. Not sure it matters, but switching to using 1 on both might be better.

fjansson: I agree, let's handle both directions in the same way. My feeling is that starting at 2 is sufficient. @CasparJungbacker, @franslql ?

fjansson: I made the i,j loops start at 2. The loops that need to end at i2,j2 for open boundaries are now only extended to i2,j2 when needed: if open boundaries are enabled and the MPI tile is at the far domain border. Frans' open boundaries test case over sea still runs.

PR #110, opened 2024-08-27
modsamptend and moddepcrossection: respect loutdirs

fjansson:
...and write per-tile output files under in subdirectories per row if loutdirs == .true. .


martinjanssens:
Ah yes of course, totally agree. But it should only do this when writing per processor, otherwise it should just write locally.

martinjanssens (approved):

PR #113, opened 2024-09-03
fix uninitialized variables

fjansson:
...and:
move a timer_toc to the end of the routine in grwdamp, and remove unused variable is_starting.

The uninitialized values are mostly namelist options missing their default values.


CasparJungbacker (approved):
LGTM

PR #114, opened 2024-09-04
genstat: do not calculate statistics for non-existing rain

CasparJungbacker:
Quick fix. Will implement something more robust in the scalar manager.


fjansson:
nsv > 0 should probably be nsv > 1, because iqr is generally 2? And you could run with a single scalar representing something else. Alternatively, the check could be if any microphysics is enabled.

PR #115, opened 2024-09-04
Dev fix genstat

fjansson:
* don't access svs field of surface scalar values (not always initialized, not needed anyway)
* calculate hur, plw, pli, clw, cli also for k=1
* don't calculate plw, pli if nsv <=1 i.e. the rain field is not present (todo: nicer condition)


CasparJungbacker (approved):
LGTM 👍

PR #116, opened 2024-09-04
Dev radtimer

lsoucasse:
Added radiation timers and fixed a bug.


fjansson (approved):
Looks good.
The previous version would read `presf_input(nlay)` which is out of range (it seems), but did not use the result. The new version doesn't read out of range, but should otherwise behave the same.

PR #117, opened 2024-09-06
Initialization from netCDF

CasparJungbacker:
As the name suggests, this PR introduces the functionality of performing cold starts from netCDF input. The variable names are a bit different from the old ASCII files, as they now follow the DALES convention.

Added a new subroutine in `modstat_nc` for reading a field (1D for now) from a netCDF file by providing the name of the variable. Essentially, it wraps `nf90_inq_varid` and `nf90_get_var`. Additionally, one can provide a default value to fill the array with in case the variable is not found.

### DIFF
--- a/src/modglobal.f90
+++ b/src/modglobal.f90
@@ -52,6 +52,7 @@ save
       integer(kind=longint) :: itrestart !<     * each trestart sec. a restart file is written to disk
       integer(kind=longint)    :: tnextrestart    !<     * each trestart sec. a restart file is written to disk
       character(50) :: startfile    !<    * name of the restart file
+      logical :: lstart_netcdf = .false.      !< Start from a NetCDF file
 
       logical :: llsadv   = .false. !<  switch for large scale forcings
       integer :: ntimedep = 100     !< maximum number of time points for time-dependent forcings
--- a/src/modnetcdf.f90
+++ b/src/modnetcdf.f90
@@ -37,12 +37,17 @@ contains
   !! the location of the error
   subroutine check(status, file, line)
 
+    use modmpi, only: myid
+
     integer,      intent(in) :: status, line
     character(*), intent(in) :: file
 
     if (status /= nf90_noerr) then
-      write(*,*) trim(nf90_strerror(status))
-      stop "NetCDF error in ", file, "on line ", line
+      if (myid == 0) then
+        write(*,*) "NetCDF error in: ", file, " on line: ", line
+        write(*,*) trim(nf90_strerror(status))
+      end if
+      stop
     end if
 
   end subroutine check
--- a/src/modnetcdf.f90
+++ b/src/modnetcdf.f90
@@ -1,0 +1,50 @@
+!> \file modnetcdf.f90
+!!  Convenience functions for working with NetCDF.
+!>
+!!  \author Caspar Jungbacker, Delft University of Technology
+!  This file is part of DALES.
+!
+! DALES is free software; you can redistribute it and/or modify
+! it under the terms of the GNU General Public License as published by
+! the Free Software Foundation; either version 3 of the License, or
+! (at your option) any later version.
+!
+! DALES is distributed in the hope that it will be useful,
+! but WITHOUT ANY WARRANTY; without even the implied warranty of
+! MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+! GNU General Public License for more details.
+!
+! You should have received a copy of the GNU General Public License
+! along with this program.  If not, see <http://www.gnu.org/licenses/>.
+!
+! Copyright 2024 Delft University of Technology
+!
+module modnetcdf
+
+  use netcdf
+
+  implicit none
+
+  private
+  
+  public :: check
+
+contains
+ 
+  !> Checks return code of netcdf calls
+  !!
+  !! If an error occurs, stops the program and print information about
+  !! the location of the error
+  subroutine check(status, file, line)
+
+    integer,      intent(in) :: status, line
+    character(*), intent(in) :: file
+
+    if (status /= nf90_noerr) then
+      write(*,*) trim(nf90_strerror(status))
+      stop "NetCDF error in ", file, "on line ", line
+    end if
+
+  end subroutine check
+
+end module modnetcdf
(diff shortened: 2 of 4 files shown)


jchylik:
> this PR introduces the functionality of performing cold starts from netCDF input.

The option to read input profiles from netCDF forcing file has been there at least for a decade (`modtestbed`, `inittestbed`). So I am wondering what is the purpose of  these commits.

Does this new option offer some additional benefits? Such as reading from _previous DALES output_?

> different from the old ASCII files, as they now follow the DALES convention

Do you mean that the subroutine uses variable names of a `profiles.XXX.nc` output?
I have had a brief look at [c563924](https://github.com/dalesteam/dales/pull/117/commits/c563924de6ce8559bc66348bc2147534f4a9c2f2) and it seems that variable names do not match.

CasparJungbacker:
@jchylik the new forcing file used in this PR acts as a replacement to both `prof.inp.XXX` and `lscale.inp.XXX`. The testbed routines you mention are too specific in the sense that they require the user to provide of additional fields.

Regarding your second point: no, this subroutine does not use the variable names from genstat output. What I meant is that the names of the input variables better match the internal DALES variables (e.g., `e12` instead of `TKE_init`.

jchylik:
I can totally see that it has some applicability for the RICO and BOMEX cases, so please do not take the following paragraphs as criticism, but rather food for thought:

>  The testbed routines you mention are too specific in the sense that they require the user to provide of additional fields.

Well, testbed now allows to bypass loading of most of the variables that are not essential. It was discussed on Zulip development planning in April, and just not pushed precisely for the reason of giving others time to express their POV.
You can see comments onwards from:
[DALES.20development.20planning/near/433307316](https://dales.zulipchat.com/#narrow/stream/382478-general/topic/DALES.20development.20planning/near/433307316)

> names of the input variables better match the internal DALES variables

Do you mean that the name of the variable of input profile (1D array) matches the name of the prognostic variable in the model (3D field)? Isn't that a bit confusing?

CasparJungbacker:
> Well, testbed now allows to bypass loading of most of the variables that are not essential

On which branch can I find those changes? If we're trying to accomplish the same here, I'm fine with just using your implementation instead. We can keep the original `modtestbed` around for compatibility. 

> and just not pushed precisely for the reason of giving others time to express their POV

Fair, but I'm not changing any existing functionality here, so nothing breaks.

> Do you mean that the name of the variable of input profile (1D array) matches the name of the prognostic variable in the model (3D field)?

Yes. Why do you think that's confusing? I think now it's more clear to see how the input maps to the prognostic variables. Though, it's mere a matter of taste. If others don't like it, we can use something else.

jchylik:
On a closer look, I can see that your code update would work well for the cases with constant forcing (modtestbed is a bit of overkill in that case.)

That said, I can see one issue in your update:
There is no checkup for the length of loaded 1D arrays.
- The subroutine `init_from_netcdf` expects the input arrays are length `kmax`, see [modstartup.f90#L589](https://github.com/dalesteam/dales/blob/dev_netcdf_init/src/modstartup.f90#L589)
- However, read_nc_field_1D read array regardless of a size, see [modstat_nc.f90#L561-L580](https://github.com/dalesteam/dales/blob/dev_netcdf_init/src/modstat_nc.f90#L561-L580)

If the length of the arrays in 'init.XXX.nc' is the same as kmax, all is fine, but if the differ, we will get segmentation fault (if we are lucky) or some very weird behaviour in the upper part of the domain.

Related to that, there is issue with the variable `height`. It is read, but never actually used. The vertical levels are still initialised by `initglobal`, see [modglobal.f90#L442-L451](https://github.com/dalesteam/dales/blob/dev_netcdf_init/src/modglobal.f90#L442-L451)

jchylik:
> On which branch can I find those changes? 

The branch is:
[to4.4.2_Fredrik_micro_testbed](https://github.com/jchylik/dales/tree/to4.4.2_Fredrik_micro_testbed "to4.4.2_Fredrik_micro_testbed") and the modified testbed module allow to bypass loading of most of the variables:
[to4.4.2_Fredrik_micro_testbed/src/modtestbed.f90](https://github.com/jchylik/dales/blob/to4.4.2_Fredrik_micro_testbed/src/modtestbed.f90).
On a second though, I should add there more bypasses with warnings (such as for loading liquid water).

fjansson:
@jchylik and @CasparJungbacker,
Jan, I forgot you had a pull request which also did something with netCDF input, sorry!

What we want now is a way to replace the ascii files for initial profiles with netCDF, in particular for a case with many scalars. Also the netCDF input should in the future work together with the scalar manager in development: scalars should be accessible by name so that you can ask for the index of the scalar for CO2 for example. Scalar profiles in the input netCDF can then be named according to what they are, instead of sv034, sv035, ...

We also discussed supporting the DEPHY format, but I guess that's not what's happening here.

CasparJungbacker:
I guess you could see this as a first step towards supporting the DEPHY standard 😉 

About @jchylik's code:

- What if the user doesn't want to do nudging? Can I provide a 1D array in `scm_in.nc` for u/v/thl etc, or does it still have to be 2D?
- I don't really like the fact that it introduces a bunch of extra switches. In my humble opinion DALES should just read whatever is given in the init file, and try to setup from that. For example: if the user provides a 1D array for u, then `ltb_u` is `.false.` implicitly, but if the user provides a 2D array the second dimension should automatically be interpreted as the "nudging dimension".

Maybe we should simply redo the startup entirely when we start working on the scalar manager 😀

jchylik:
> What we want now is a way to replace the ascii files for initial profiles with netCDF, in particular for a case with many scalars. 

Yes, that makes sense. As mentioned above, both options can easily coexist:

-  `init_from_netcdf` for cases with simple forcings
-  `modtestbed` subroutines for cases with time-dependent forcings, and other perks (initialisations of cloud droplets and ice,  altitude dependent nudging, ...)


Following on that:

>  if the user provides a 1D array for u, then ltb_u is .false. implicitly, but if the user provides a 2D array the second dimension should automatically be interpreted as the "nudging dimension"

Well, sometimes we just want to nudge to a constant profile. And in many cases we do not want to nudge in the whole domain, but only adaptively above certain altitude (such as as strongest inversion above the residual layer). 

> We also discussed supporting the DEPHY format, but I guess that's not what's happening here.

The development of modtestbed is moving in that direction, but for a backward compatibility, we would still keep also the option to load by old variable names.  

> we should simply redo the startup entirely 

Well, if we are adjusting modstartup, then we should also change modglobal.

PR #121, opened 2024-09-17
Higher order advection schemes for open boundary conditions

franslql:
Added the possibility to use 5th and 6th order advection schemes in combination with open boundary conditions.


fjansson (approved):
Possibly a small surprise in modadvection:
advec*_2nd_boundary_buffer changes the istart,iend,jstart,jend arguments,
so that after the call they describe the inner part of the domain.

PR #126, opened 2024-10-03
Added surface stresses as rho*ustar**2

franslql:


fjansson (commented):
Thanks, looks good to me in general.

* I'd suggest `**2.` -> `**2` because raising to a non-integer power is more expensive (the compiler might be clever and change back to int, but I wouldn't bet on it).

* It's not immediately clear to me how to split rho*ustar**2 into the the two directions i.e. `upcu2/horv2`. Do you have a reference or reason for the square in the direction factor?

* It might be good to add some small epsilon to horv2 to avoid 0/0. It's done in other places in the surface scheme for that direction factor.

franslql:
> Thanks, looks good to me in general.
> 
> * I'd suggest `**2.` -> `**2` because raising to a non-integer power is more expensive (the compiler might be clever and change back to int, but I wouldn't bet on it).

Sounds good. I took It from the modsurface routine.

> * It's not immediately clear to me how to split rho*ustar**2 into the the two directions i.e. `upcu2/horv2`. Do you have a reference or reason for the square in the direction factor?

In modsurface in the calculation for dudz and dvdz (line 881 and 882 for example) ustar is multiplied by u/uhor and v/uhor respectively. The square comes from squaring; (ustar*u/uhor)**2.

> * It might be good to add some small epsilon to horv2 to avoid 0/0. It's done in other places in the surface scheme for that direction factor.

Yes that might be a good idea.

fjansson:
I'll merge this, and change the **2 and add some epsilon afterwards. Thanks for the explanation, dudz and dvdz is a good reference.

PR #128, opened 2024-10-18
New tracer manager

CasparJungbacker:
New features:
- Providing tracer attributes and profiles via NetCDF.
- To enable the previous, the changes introduced by #117 have been included in this branch. Intialization is still a point of discussion so this should be seen as a prototype.
- "Internal tracers", like `qr` and `Nr`, no longer have to be provided in `scalar.inp.XXX`.
- User no longer has to provide `nsv` in the namelist. DALES figures out itself how much memory should be allocated based on the amount of internal tracers and the amount of tracers provided in `tracers.XXX.nc`.

Removed:
- The ability to specify the advection scheme per tracer. `iadv_sv` is now a single value, rather than a list.

TODO:
- [x] Documentation on how to use the new input.


CasparJungbacker:
@fjansson alright, I think that's all. Regarding the DEPHY input names, `dqtdxls` and `dqtdyls` don't seem to have a DEPHY-equivalent (https://github.com/GdR-DEPHY/DEPHY-SCM/blob/master/DEPHY-SCM_CommonFormat_v1.0.pdf), so I left them alone.


Maybe also important to note: I've tried to keep these changes as backwards-compatible as possible, but the changes to the `iadv_sv` switch did break the cases where the `iadv_sv` is provided as a list in the namelist (i.e.: `iadv_sv=100*7`)

fjansson (approved):
Nice! Mostly backwards compatible but adds support for initial profiles in .netcdf, DEPHY format.

PR #131, opened 2024-10-25
Fix Intel build in CI (once again)

CasparJungbacker:


fjansson (commented):
Apparently intel is deprecating the old `ifort` compiler in favor of the new `ifx`.
https://github.com/viperleed/viperleed/issues/189

Maybe we could add another system `localpc_ifx` and put `mpiifx` there?
Then we don't keep the the old ifort name hanging around pointing to the new compiler, and we also keep some backwards compatibility because there may be some users of ifort still.

CasparJungbacker:
Sounds good. Do you want to keep around the old entries that use `mpiifort`? I think the presets that use `mpiifort` are for systems that no longer exist anyway, so I could clean the CMake file up a bit at the same time.

fjansson:
I would keep `localpc_ifort`, the entries for specific machines that no longer exist can be removed.

PR #132, opened 2024-10-28
Fix compiler warings, mainly unused variables

fjansson:
372 warnings less.
* changed line endings of modtracers DOS->Unix
* did not touch modbulkmicroc sine it's being worked on
* did not touch RRTMG*


CasparJungbacker (approved):
Nice, the compilation output is noticeably cleaner. A lot of the left-over warnings seem to come from variables related to time-averaging (real to integer), so I'm wondering what the best way to do this explicitly is.

CasparJungbacker (approved):

PR #133, opened 2024-11-01
Enable FFTW by default

CasparJungbacker:
Since most systems have FFTW installed anyway:
- Set USE_FFTW to ON
- Set solver_id to 100 when USE_FFTW is defined

FFTW can still be disabled by setting `-DUSE_FFTW=False`.

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -22,7 +22,7 @@ jobs:
       fail-fast: false
       matrix:
         toolchain: ["gnu", "intel"]
-        fftw:    ["", "-DUSE_FFTW=T"]
+        fftw:    ["-DUSE_FFTW=F", "-DUSE_FFTW=T"]
         hypre:   ["", "-DUSE_HYPRE=T"]
         
     # Steps represent a sequence of tasks that will be executed as part of the job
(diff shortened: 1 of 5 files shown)


CasparJungbacker:
@fjansson looking at the compilation CI pipeline, should we also remove all tests with `-DUSE_FFTW=True`?

fjansson:
we could remove them but instead add at least some test with `-DUSE_FFTW=False`

fjansson:
To keep the number of compilation tests down, I'd say we don't need to test all combinations with -DUSE_FFTW=False. Testing one could be nice... but more important is that DALES still runs with the old FFT.

CasparJungbacker:
Ah, saw your comment right after I enabled tests for `-DUSE_FFTW=F` 😂 I'll revert that one, and add a test run of BOMEX with FFTW disabled then.

CasparJungbacker:
Upon closer inspection, I saw this in the testing configuration:

```python
@pytest.mark.parametrize("solver", [0, 100]) # FFTPACK & FFTW
```

meaning that we've been testing for both the included FFT lib and FFTW for some time now 👍

fjansson (approved):

PR #134, opened 2024-11-04
modradrrtmg: merge duplicate definitions of orbital parameters

fjansson:
Some orbital parameters were defined both in modradrrtmg and modraddata. Moved all orbital parameters to modraddata, which is where shr_orb_decl() looks for them. This is to resolve use of undefined variables found on Fugaku (debug build). Note the orbit calculations aren't actually used anyway.

### DIFF
--- a/src/modraddata.f90
+++ b/src/modraddata.f90
@@ -155,12 +155,14 @@ SAVE
                                                      ssaaersw, &
                                                      asmaersw, &
                                                      ecaersw
-  real(SHR_KIND_R4)                             :: eccen,     &  ! Eccentricity
-                                                   obliqr,    &  ! Earths obliquity in radians
-                                                   lambm0,    &  ! Mean long of perihelion at the vernal equinox (radians)
-                                                   mvelpp,    &  ! moving vernal equinox longitude of perihelion plus pi (radians)
-                                                   delta,     &  ! Solar declination angle in rad
-                                                   eccf          ! Earth-sun distance factor (ie. (1/r)**2)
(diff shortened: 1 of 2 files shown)


CasparJungbacker:
Another reason to explicitly pass variables as arguments to subroutines :^). 

Looks good, but when you say "the orbit calculations aren't actually used", do you mean that they're just not used by RRTMG, or _at all_?

fjansson:
I mean at all, at least that's what I believe but a double check would be good. There's somewhere a much simpler orbit calculation which is used instead, with the assumption that the orbit is a perfect circle. So the times of sunrise, sunset are off by up to ~30 min. 
And yes, this is a case where function arguments would be great.

CasparJungbacker:
Hmmm, we have `setupSW()` in both RRTMG and RRTMGP, which both call `shr_orb_decl()` from `shr_orb_mod`, so at least it's called. Whether the call actually does something is a different question, I think the values it touches in `modraddata` are not used at all.

CasparJungbacker (approved):

fjansson:
the setup is [called](https://github.com/dalesteam/dales/blob/92b146a0ae4bced7dfba8213f2dca83547575a83/src/modradrrtmg.f90#L875), but then to actually calculate the zenith angles, the function [zenith](https://github.com/dalesteam/dales/blob/92b146a0ae4bced7dfba8213f2dca83547575a83/src/modraddata.f90#L216) in modraddata is used, and that one seems to not care about the fancy orbit parameters.

We should fix that too, but let's do another PR for it, and check the zenith angles with some online astronomic calendar.

PR #135, opened 2024-11-06
SB/KK microphysics cleanup

CasparJungbacker:
# Changes
In preparation for aerosols, some cleanup of the warm microphysics.

- Split SB and KK schemes into separate modules, similar to how the different advection schemes are implemented. This has several benefits:
  - The implementations of different sources are much more readable, as the subroutines are shorter.
  - Prevents using SB parameters in the KK scheme and vice-versa.
  - More modular approach makes it easier to add new schemes.
- Added more `field_r`'s. When using FP32, all computations should now actually be done in FP32.
- Some (opinionated) formatting.

# Tests
Used the benchmnark/rico case for testing. Started from a 24 hour warm start, then simulated another hour. Here are some plots of the different `qr` tendencies:

image

image

I think it would be a good idea to have someone who knows the microphysics go over the parameters, since modmicrodata has become very messy over the years.

@fjansson anything else that should be done?


fjansson:
The curves look good.

This, in `erfint` 
```
erfymin = erf(abs(ymin))
erfymax = erf(abs(ymax))
erfymin = sign(erfymin, ymin)
erfymax = sign(erfymax, ymax)
```
can be simplified to
```
erfymin = erf(ymin)
erfymax = erf(ymax)
```
since [erf](https://en.wikipedia.org/wiki/Error_function) is antisymmetric, right? I think the abs and sign are remains of the previous polynomial expansion which may not have been antisymmetric. But the intrinsic erf should be. Anyway, I think this is used only if `l_lognormal` is true, which is not the default.

fjansson (approved):

PR #141, opened 2024-11-26
CMake fixes

CasparJungbacker:
Remove CMake presets again :-)

Now, select a Fortran compiler using regular CMake variables. E.g., for GNU:
```
FC=mpif90 cmake <options> <path>
```
and for Intel:
```
FC=mpiifx cmake <options> <path>
```


fjansson (approved):
Still to do: to set a default CMAKE_RELEASE_TYPE. 
Tested on Fugaku and on Arch Linux.

PR #146, opened 2024-12-11
DEPHY input for nudging

CasparJungbacker:
This PR adds the possibility of reading nudging profiles and nudging constants from a NetCDF input file in DEPHY format. Nudging constants can now be provided on a per-variable basis, allowing different constants for different variables. Also, did some cleanup, as always :)

Still need to test this a bit further with a case that actually uses nudging. Perhaps botany?

I also plan to implement nudging for tracers, but will make a separate PR for that.


CasparJungbacker:
@fjansson I've tested the new setup for botany and it seems to run fine. From my part you can merge if you're fine with it too.

fjansson (approved):

PR #147, opened 2024-12-20
Fix uninitialized variables

fjansson:
* Uninitialized variables in the new LSM: two actually uninitialized and some more are now zeroed before reading in netcdf, to avoid uninitialized halos.

* in the open boundary conditions, phase velocity was used uninitialized in the first time step

### DIFF
--- a/src/modopenboundary.f90
+++ b/src/modopenboundary.f90
@@ -43,8 +43,9 @@ contains
   subroutine initopenboundary
     ! Initialisation routine for openboundaries
     use modmpi, only : myidx, myidy, nprocx, nprocy, myid
-    use modglobal, only : imax,jmax,kmax,i1,j1,k1,dx,dy,itot,jtot,solver_id,nsv,cu,cv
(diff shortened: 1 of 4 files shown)


src/modopenboundary.f90
```diff
@@ -177,6 +178,7 @@ subroutine openboundary_initfields
     if(.not.linithetero) return
     u0 = 0.; um = 0.; v0 = 0.; vm = 0.; w0 = 0.; wm = 0.;
     thl0 = 0.; thlm = 0.; qt0 = 0; qtm = 0; e120 = 0.; e12m = 0.
+    up = 0.; vp = 0.; wp = 0. ! do we need this?
```
CasparJungbacker: `modfields/initfields` already zeroes these arrays, and is called before `initopenboundary`

CasparJungbacker (approved):

src/modopenboundary.f90
fjansson: thanks! I'll remove that line and then merge

PR #152, opened 2025-01-24
modbulkmicro: do clipping over all vertical levels

fjansson:
On GPU, a loop range to qrroof doesn't seem to work if qrroof is 0
- complains about data not present. Also on CPU, it's better to clip all levels since advection could introduce negative qr.

### DIFF
--- a/src/modbulkmicro.f90
+++ b/src/modbulkmicro.f90
@@ -315,21 +315,8 @@ module modbulkmicro
 
     call bulkmicrotend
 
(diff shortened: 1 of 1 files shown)


CasparJungbacker:
Does it make sense to also apply this change to line 332?

CasparJungbacker (approved):

fjansson:
yes, definitely. And the early return before that loop should also go. I'll do it.

PR #154, opened 2025-02-06
Tracer manager updates

CasparJungbacker:
Main changes:
- Fixed handling of ASCII input files for tracer profiles and properties. The user provided tracers are inferred from the column names in `scalar.inp`. Properties are read from `tracerdata.inp`.
- Better printing during startup (only root PE has to print)
- Kinematic surface flux for tracers is read from `tracerdata.inp` instead of namoptions

Smaller changes:
- Added new function `print_info_stderr` to `modmpi`. This is a helper function that prints a message to stderr, but only for the root PE. I.e., it does the check for `myid==0`.
- Introduced the global variable `iinput` for selecting between input file formats.
- Added `modlsmdata.f90` that contains the LSM variables. This was needed to resolve a circular dependency


fjansson (approved):
tested several combinations of ASCII input.

PR #156, opened 2025-02-07
Dev pressurefix; horizontal pressure gradient

StevenvdLinden:
1. Added new option for horizontal pressure gradient (to enable pressure-driven channel flow). Also supports time-dependent pressure gradient.
2. `lcoriol` and `lpressgrad` switches are now mutually exclusive, fixing **implicit** behavior that pressure gradient would always be set to zero, when `lcoriol=.false.` (see modforces.f90, lines 370-379). 
3. Input infrastructure remains the same. Columns in *lscale.inp.iii* are read in as either geostrophic winds (when `lcoriol=.true.` **or** as horizontal pressure gradients when `lpressgrad=.true.`. Same for netcdf input.
4. Error is produced when both are set to true.

**Background**: the old switch `lpressgrad` would have allowed to impose no pressure gradients, whilst still allowing Coriolis forces. This option is now made explicit: you must put geostrophic winds to zero when you want this to happen (as intuitively most of us would have already done anyway)


fjansson:
I like the proposed change, to bring back the pressure gradient way of running a case. The commits are complicated though - there are some unrelated changes too, possibly as the result of a merge conflict. I will try to gather the changes in a minimal commit.

Then the pressure gradient option needs to be documented along with the other large scale forcings.

I've got some general wishes (could go in the manual if we agree):
* If you have your fork of the main repository, make sure the branch you are basing your work on (dev in this case) is in sync with the main repository before you start the change. There is a button "Sync Fork" on the github page of your fork, probably you have to select the dev branch and then press it. It looks like the commits here are based on something quite old and that may be the reason for the complicated merge.
* for comments, it's generally enough if they describe the code as it is now. Description of changes and when they were made is kept track of in the git history anyway, and on a higher level in the change log with new releases. If we accumulate them in the source code itself, it becomes harder to read and manage. 
* avoid mixing model changes and code formatting changes - especially formatting changes in unrelated parts of the code
* avoid whitespace at the end of the lines (code editors should have a function to remove these).

and one more, not applicable in this case:
* don't indent with tabs (not Fortran standard, editors are not consistent with tab width anyway)

fjansson:
First attempt in https://github.com/dalesteam/dales/tree/dev-pressuregradient
@StevenvdLinden can you see if I missed something? modstartup had the most manual work, there's a fair chance I lost something there. The other files I did by cherry-pick + light editing afterwards, they have a good chance of being as you wanted them.  

Testing: it currently compiles, and it runs an old RICO case with ug,vg without obviously breaking. Please try a case with lpressgrad = true.

StevenvdLinden:
Fixed PR (merged up-to-date dev into PR), reworked some comments

StevenvdLinden:
Clean implementation of pressure gradients carried over in branch dev-pressuregradient

PR #159, opened 2025-02-17
modmicrophysics: stop and warn if Nc_0 is unreasonably small

fjansson:
The cloud droplet number concentration Nc_0 should be given in number per m3. Giving it in the common unit of number per cm3 results in a confusing and hard to track down error from RRTMG at dawn (ask me how I know):

LIQUID SSA GRTR THAN 1.0

This commit adds a stop with a warning, if Nc_0 < 1e6 at initialization,  which is a reasonably sure sign it was given in the wrong unit.

There is probably a bug in RRTMG as well - it says it accepts cloud droplet sizes up to 60 um, but a droplet > 59 um turns what should be an interpolation in a table into an extrapolation, and the extrapolated value exceeds the limit 1.0, which is checked for afterwards.

### DIFF
--- a/src/modmicrophysics.f90
+++ b/src/modmicrophysics.f90
@@ -66,7 +66,15 @@ contains
             ! Bulkmicro 3 special treatment of saturation adjustment is currently only impmemented in the fast_thermo scheme
          end if
       end if
-    end if
+
+      if (Nc_0 < 1e4) then
+         ! Check that Nc_0 is reasonable.
+         ! It's easy to give it in units of /cm3 by mistake,
+         ! and when run with RRTMG that results in a hard-to-understand crash
+         ! in the short-wave scheme
+         STOP 'Nc_0 is suspiciously small (unit should be number per m3).'
+      end if
+   end if
 
     call D_MPI_BCAST(imicro,   1, 0,comm3d,ierr)
     call D_MPI_BCAST(l_sb,     1, 0,comm3d,ierr)


fjansson:
It seems CI broke, but I it doesn't look like the one commit here caused it?

CasparJungbacker:
For some reason the runner is checking out v1.6 of RRTMGP, while we're using v1.8. Did you maybe change the `.gitmodules` file? In the change log of this PR it says that the submodule was updated.

fjansson:
I see it now! I don't know how I did that, certainly not on purpose. Nice catch by the CI, I guess. I'll fix it and commit again.

jchylik:
> This commit adds a stop with a warning, if Nc_0 < 1e6 at initialization, which is a reasonably sure sign it was given in the wrong unit.

Is that a sure sign? 
What if we want to simulate the effect of trubulence on radiative fog?
Do you really think it requires STOP instead of just warning?

Not to mention there are cases of very low cloud droplet numbers in Arctic and Antarctic air.

fjansson:
@jchylik what's the smallest Nc you can imagine? I'm fine with decreasing the threshold. I think warnings are easily lost among all the other output, so if it's not going to run well anyway  we might as well stop.

jchylik:
> @jchylik what's the smallest Nc you can imagine?

Not sure, but I can look that up.

> I think warnings are easily lost among all the other output, so if it's not going to run well anyway we might as well stop.

Good point. 

Still, it would be good practise to set `parameter Nc_0min` in modmicrodata.f90 instead of hard-coding value in the middle of a long module.

CasparJungbacker:
> The cloud droplet number concentration Nc_0 should be given in number per m3. Giving it in the common unit of number per cm3 results in a confusing and hard to track down error from RRTMG at dawn (ask me how I know):
> 
> LIQUID SSA GRTR THAN 1.0

Curious, I've never seen this error appear in any of my runs with interactive aerosols. Even though in those cases you'd have plenty of cells where `Nc(i,j,k)=0`. For that reason I'm also a bit hesitant on implementing a hard lower bound on `Nc`, but I guess I can implement a workaround for when aerosols are enabled.

fjansson:
The check I added is only done at initialization, on the Nc_0 specified in the namelist, so should not interfere with the aerosol scheme. I basically just want a sanity check on the input to catch the case where the unit is wrong. Per cm3 vs m3 is a factor 1 million, so I think we can find a safe limit that will not harm any run.

fjansson:
I changed the limit to 1e4 per m3 = 0.01 per cm3.

CasparJungbacker (approved):

CasparJungbacker:
Maybe we could even add a (hidden) flag to overwrite such stops? For those that know what they're doing

fjansson:
I thought about it, didn't want to go there (yet?). I'll merge like this for now, if someone hits the limit with a small Nc set on purpose we can revert it or add the I-know-what-I'm-doing flag.

PR #160, opened 2025-02-20
Fix double normalization of cfrac vertical profile

fjansson:
cfrac was divided by ijtot twice, removing one of them, resulting in very small cloud fractions in the vertical profile.
Also remove unused variables in modgenstat.


martinjanssens (approved):
Nice, thanks very much for the fix! This looks perfect now as far as I can see :)

PR #161, opened 2025-02-26
Update simpleice, bring it closer to simpleice2

fjansson:


src/modsimpleice.f90
```diff
@@ -131,6 +132,8 @@ subroutine simpleice
     integer:: i,j,k
     real:: qrsmall, qrsum,qrtest
 
+    call timer_tic('modsimpleice', 1)
+    call timer_tic('modsimpleice/setup', 1)
```
CasparJungbacker: Just a suggestion: I've recently started giving every module and subroutine a character parameter for their name. This makes calling the timers easier and less error-prone. E.g:

```fortran
module modsimpleice
  character(len=*), parameter :: modname = "modsimpleice"
contains
  subroutine simpleice
    character(len=*), parameter :: routine = modname//"/simpleice"
    call timer_tic(routine, 1)
    ...
    call timer_toc(routine) <--- automatically the correct argument
  subroutine simpleice
end module modsimpleice
```

martinjanssens (approved):

PR #162, opened 2025-03-11
Update RTE-RRTMGP to version 1.9

CasparJungbacker:
- Use effective diameter of ice particles for cloud optics
- Add RRTMGP-data repository as submodule

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -11,3 +11,6 @@
 	path = external/ecbuild
 	url = https://github.com/ecmwf/ecbuild.git
 	branch = develop
+[submodule "external/rrtmgp-data"]
+	path = external/rrtmgp-data
+	url = git@github.com:earth-system-radiation/rrtmgp-data.git
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -183,16 +183,17 @@ set( rrtmgp_srcs
 	${PROJECT_SOURCE_DIR}/external/RTE-RRTMGP/rte-kernels/accel/mo_rte_solver_kernels.F90
 	${PROJECT_SOURCE_DIR}/external/RTE-RRTMGP/rte-kernels/mo_rte_util_array.F90
 	${PROJECT_SOURCE_DIR}/external/RTE-RRTMGP/rte-kernels/mo_fluxes_broadband_kernels.F90
+	${PROJECT_SOURCE_DIR}/external/RTE-RRTMGP/rte-kernels/mo_rte_kind.F90
 	${PROJECT_SOURCE_DIR}/external/RTE-RRTMGP/rte-frontend/mo_fluxes.F90
 	${PROJECT_SOURCE_DIR}/external/RTE-RRTMGP/rte-frontend/mo_optical_props.F90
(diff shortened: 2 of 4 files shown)


martinjanssens (commented):
Nice, thanks Caspar. I think the error is that RRTMGP always made it look like it needs ice effective radius as input to its lookup tables, when really it needed effective diameter (see e.g. the new cloud optics: https://github.com/earth-system-radiation/rte-rrtmgp/blob/f05e93419806694bcf58641f2ebd0065d28dbefb/rrtmgp-frontend/mo_cloud_optics_rrtmgp.F90). I think we are now giving RRTMGP effective radius, here:
https://github.com/CasparJungbacker/dales/blob/8c82b9dbeb6eeb2b8e271e200a1ef256499b5a12/src/modradrte_rrtmgp.f90#L590
And I think (but correct me if you think it is passed to RRTMGP otherwise), we'd resolve the issue by multiplying the number calculated there by 2. For legibility/clarity, we might want to actually change the `reIce` variable in `modradrte_rrtmgp` to `deIce` or similar. What do you think?

martinjanssens:
Incidentally, I believe that for the liquid water LUTs, we *should* pass the effective *radius* (like we are currently doing) and not diameter, which seems to have made the whole thing confusing for a while.

CasparJungbacker:
Looking at the RRTMGP frontend, I think it still expects an effective *radius*:

```fortran
  function cloud_optics(this,                     &
                        clwp, ciwp, reliq, reice, &
                        optical_props) result(error_msg)
    class(ty_cloud_optics_rrtmgp), &
              intent(in   ) :: this
    real(wp), intent(in   ) :: clwp  (:,:), &   ! cloud liquid water path (g/m2)
                               ciwp  (:,:), &   ! cloud ice water path    (g/m2)
                               reliq (:,:), &   ! cloud liquid particle effective size (microns)
                               reice (:,:)      ! cloud ice particle effective radius  (microns)
```

meaning that we're doing it correctly on our side. However, `cloud liquid particle effective size` confuses me...

martinjanssens:
I agree that this is still confusing. Still I'm pretty sure we need to multiply by 2. We're indeed calling that `cloud_optics` function you're referencing above, but it just calls `compute_cld_from_table` without any multiplication by 2, i.e. it still just goes to the lookup table, which is tabulated by diameter for ice. Maybe they missed this when relabeling just `radIce` (https://github.com/earth-system-radiation/rrtmgp-data/issues/8)? We could double check if you want.

CasparJungbacker:
Ahaa, you're right. Indeed, it looks like they forgot about updating the comments in the subroutine.

CasparJungbacker:
So, to double check, what we should be doing is `iceRe = iceRe * 2`, right @martinjanssens ?

martinjanssens:
Yes, exactly!

CasparJungbacker:
Ok, I think that should be it then!

martinjanssens:
That looks good! Although I'm not 100% sure if the limiters (10, 180) should also be applied in diameter, rather than radius. Let me check.

martinjanssens:
Yes, the limiters are also in terms of diameter (they did update the names of those to `diamice_lwr` and `diamice_upr` in the lookup tables), so we should convert to diameter before comparing to those limiters. Otherwise it looks ready!

PR #163, opened 2025-03-18
Save bulkmicrophysical tendencies per processor and add option to save ql flux in samptend

MarloesvD:


fjansson:
Closing this PR, because it was implemented in #172 and #174.

PR #166, opened 2025-04-02
Dev emissions

adoyenne:
Modtracers and openboundaries: add tracer manager in openboundary to read boundary conditions by their name and modtracer_type is created to avoid cycling dependancies
Emissions: point sources emission input and processing (plume rise parameterization) has been added
LSM:  fix uninitialized fields

### DIFF
--- a/src/modemisdata.f90
+++ b/src/modemisdata.f90
@@ -32,7 +32,7 @@
 ! 'location', i.e. switch scalar field represents which species?
 
 module modemisdata
-  
+
   implicit none
   save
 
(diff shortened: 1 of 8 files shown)


src/modemisdata.f90
```diff
@@ -32,25 +32,28 @@
...
+              npoints  =  0        ! no. of point sources in each MPI process file
+
   logical  :: l_scale = .false.    ! emission scaling switch
   real, dimension(100) :: scalefactor = 1
 
-  character(len = 6), dimension(100) :: & 
+  character(len = 6), dimension(100) :: &
               emisnames = (/ ('      ', iname=1, 100) /) ! list with scalar names,
-                          ! each name must(!) be 6 characters long for now  
+                          ! each name must(!) be 6 characters long for now
```
CasparJungbacker, line 56: Should we make this a property of the Tracer type instead? I.e., add an attribute `lemis` to each tracer in `tracers.XXX.nc`?

CasparJungbacker (approved):

src/modemisdata.f90
adoyenne: Hi Caspar, Yes, it can be a property of the Tracer type, I think, if this will not violate anything in the emission module

fjansson (approved):

PR #167, opened 2025-04-03
Read skin temperature from input netcdf file

ChrisC28:
I've modified modsurface.f90 to 
- read from a netcdf file with dimension x,y,time
- initialise profiles from input tskin
- temporally interpolate to the current model time step (following the proceedure in the temporally varying surface flux routines)

I've added several helper subroutines in modnetcdf.f90 to faciliate read write.
The user needs to specify ltskininp = .true. in the NAMOPTIONS file. 

Input checking needs some additional work. For example, need to check the input file timesteps to ensure that they align with the model timesteps. At present, the model will not fail but simply set the tskin equal to the fist timestep in the file.

### DIFF
--- a/src/modstartup.f90
+++ b/src/modstartup.f90
@@ -364,6 +364,9 @@ contains
     call initradiation
     call initchem
     call initsurface
+    PRINT *, thls
+    PRINT *, '=============' 
+
     call initdatetime
     call initemission
     call initlsm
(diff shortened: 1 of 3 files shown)


src/modstartup.f90
```diff
@@ -364,6 +364,9 @@ subroutine startup(path)
     call initradiation
     call initchem
     call initsurface
+    PRINT *, thls
+    PRINT *, '============='
```
CasparJungbacker, line 368: I guess these print statements are for debugging? Can you remove those before we merge your branch? Otherwise, the changes look good to me :-)

CasparJungbacker (changes requested):

PR #168, opened 2025-04-07
per-processor microphysics tendencies

fjansson:
Add the option to output microphysics tendencies per MPI domain.
This is one part of  PR #163 by Marloes van Driel.

Sample namelist to test:
```
&NAMGENSTAT    
lstat      = .true.
dtav       = 300
timeav     = 300  
/
&NAMBULKMICROSTAT
lmicrostat = .true.
dtav = 300     
timeav = 300
/
&NAMSAMPLING
lsamptend = .true.
lprocblock = .true.
dtav = 300    
timeav = 300  
/
```

### DIFF
--- a/src/modbulkmicrostat.f90
+++ b/src/modbulkmicrostat.f90
@@ -47,12 +47,12 @@ save
   integer(kind=longint):: idtav, itimeav, tnext, tnextwrite
   integer          :: nsamples
   logical          :: lmicrostat = .false.
-  integer, parameter      :: nrfields = 6  , &
-                 iauto    = 2 , &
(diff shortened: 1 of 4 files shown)


fjansson:
I've tested this on RICO with and without per-processor stats enabled. @CasparJungbacker can you see if it compiles with OpenACC - it probably won't run and I'm fine with that for now, I just want it to not break the compilation.

MarloesvD:
I think I found a small mistake, shall I do a new PR after this one is merged?

CasparJungbacker (approved):
It compiles just fine with OpenACC!

PR #179, opened 2025-04-25
Refactor microphysics

CasparJungbacker:
Clean up and preparation for the upcoming aerosol module.

Main changes:
- Compute statistics per-process with the new statistics module. This makes `modbulmicrostat` redundant, but `modsimpleice` still uses it so I left it in for now.
  - This change also introduced a bunch of extra temporary fields, need a solution for that at some point (some memory pool module?) (to tackle in esiwace?).
- Got rid of arrays for xr, lbdr, mur, dvr, qrmask, qcmask. All of them are computed in-place now -> saves 6 3D fields :-)
- More strict separation of the different schemes
- Added `modnamelist`

Topic for discussion: we still have the code for the option `l_lognormal=.true.`, but this option is read from the namelist, so there's currently no way to enable it. Should we remove the option all together, or add it to the microphysics namelist again?


CasparJungbacker:
Running some tests now...

CasparJungbacker:
Some profiles after 1 day of RICO + ~1 hour:

image

Good enough @fjansson?

fjansson:
Looks quite good!
another way to compare the two schemes could be to compare the microphysics tendency statistics. That would be a more direct comparison of the outputs of the two microphysics scheme. We can do that also after merging this.

fjansson (approved):

PR #183, opened 2025-05-13
Dev fix ibm

fjansson:
fix slabavg range for scalars in initialization, and better error messages when IBM namelist reading fails.

### DIFF
--- a/src/modibm.f90
+++ b/src/modibm.f90
@@ -59,7 +59,8 @@ contains
                                fname_options, nsv, cu, cv, ijtot, &
                                iadv_mom,iadv_tke,iadv_thl,iadv_qt,iadv_sv,iadv_cd2, &
                                ibas_prf, &
-                               dx,dy,fkar
+                               dx,dy,fkar,&
+                               checknamelisterror
     use modmpi,         only : myid, comm3d, mpierr, myidx, myidy, d_mpi_bcast, excjs, D_MPI_ALLREDUCE, &
                                mpi_max, mpi_sum
     use modsurface,     only : lmostlocal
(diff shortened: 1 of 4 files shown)


src/modibm.f90
```diff
@@ -85,13 +86,7 @@ subroutine initibm
...
       open (ifnamopt,file=fname_options,status='old',iostat=ierr)
       read (ifnamopt,NAMIBM,iostat=ierr)
-
-      if (ierr > 0) then
-        print *, 'Problem in namoptions NAMIBM'
-        print *, 'iostat error: ', ierr
-        stop 'ERROR: Problem in namoptions NAMIBM'
-      end if
-
+      call checknamelisterror(ierr, ifnamopt, 'NAMFIELDDUMP')
```
StevenvdLinden: there is a small typo here: 'NAMFIELDDUMP' should be changed into 'NAMIBM'

StevenvdLinden (commented):
Hi Fredrik. Changes look good. There is one small mistake in the checknamelist call, where you refer to NAMFIELDDUMP istead of NAMIBM

src/modthermodynamics.f90
```diff
@@ -396,7 +396,7 @@ subroutine diagfld
     call slabavg(qt0,fluid_mask,ih,qt0av)
     call slabavg(ql0,fluid_mask,ih,ql0av)
     do n=1,nsv
-      call slabavg(sv0(:,:,:,n),fluid_mask,ih,sv0av(1:1,n))
```
StevenvdLinden, line 399: addition: this original mistake also seems to occur in modstartup. Can you correct it there as well please?

src/modibm.f90
fjansson: thanks for finding that, it's a copypaste mistake. I'll fix it.

PR #186, opened 2025-05-20
Dev emissions

adoyenne:
Hi,

2 commits:
1. modemission: point source input change, explicit plume rise option
2. modlsm: add optional Ags model, CO2 ecosystem fluxes, resistances


fjansson (approved):
things to check: 
is it safe to communicate derived data types with MPI?

PR #188, opened 2025-05-28
Add spraying module

fjansson:
Add the possibility to have a source of a scalar and liquid water spray in the model.
The sprayed scalar can be nudged to 0 at the boundary, with a sponge layer.
Implemented by Stephan de Roode, Pim van Dorp

limitations:
- at most one source in the domain
- ...emitting one scalar
advantages:
- easy to use, the source is described with a namelist entry. No input files needed.

I see this as somewhat experimental, we may want to change the settings or merge this with the point emissions module.

Wishes:
- more sources
- configure which scalars are emitted
- configure which scalars are nudged to 0 at the boundary

Usage:
```
&NAMSPRAYING
lwater_spraying = .false.
lsalt_spraying  = .true.
i_glob_spray = 102
j_glob_spray = 119
k_glob_spray = 3
water_spray_rate = 1e-20
salt_spray_rate  = 1
lsalt_sponge = .true.
/
```

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -80,6 +80,8 @@ set( dales_srcs
 	modsimpleice.f90
 	modsimpleice2.f90
 	modslabaverage.f90
+	modspraying.f90
+	modsprayingdata.f90
 	modstartup.f90
 	modstat_nc.f90
 	modstat_profiles.f90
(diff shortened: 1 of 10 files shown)


CasparJungbacker:
Is this the same spraying module Annelot has been using?

fjansson:
Yes, it's the same one. For that reason it could be worth to merge now even if we may want to unify with the other emission module later.

CasparJungbacker (approved):

PR #198, opened 2025-07-30
OpenACC for simpleince and modsurface(lmostlocal=True)

fjansson:
Add OpenACC directives to simpleice - seems to work, not seriously tested

Also add one missing directive in modsurface - gives a weird warning
`NVFORTRAN-W-1054-Module variables used in acc routine need to be in acc declare create() `
and then fails when linking with `Undefined reference to 'modsurface_patchynr_' in 'CMakeFiles/dales.dir/modsurface.f90.o'` @CasparJungbacker any idea how to fix that? Otherwise we can live without patch support for now.

### DIFF
--- a/src/modsurface.f90
+++ b/src/modsurface.f90
@@ -181,6 +181,8 @@ contains
     call D_MPI_BCAST(ltskininp                  ,            1, 0, comm3d, mpierr)
     call D_MPI_BCAST(min_horv                   ,            1, 0, comm3d, mpierr)
 
+    !$acc update device (xpatches, ypatches)
+
     if(lCO2Ags .and. (.not. lrsAgs)) then
       if(myid==0) print *,"WARNING::: You set lCO2Ags to .true., but lrsAgs to .false."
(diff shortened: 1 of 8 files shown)


src/modsurface.f90
```diff
@@ -1675,6 +1675,7 @@ function factorial(k)
   end function factorial
 
   function patchxnr(xpos)
+    !$acc routine seq
     use modmpi,     only : myidx
     use modglobal,  only : imax,itot
```
CasparJungbacker: I think the issue here is that `patchxnr` and `patchynr` use global variables. Hence, these need to be copied to the GPU explicitly by using `!$acc declare create`. So for example, we would need to place an `!$acc declare create(imax, itot)` near the declaration of both of them. That being said, I think the more elegant solution here is to pass everything as arguments to the functions.

fjansson: I added !$acc declare create() for all the variables needed, to mess as little as possible with the patch{x,y}nr functions.

PR #205, opened 2025-08-17
preventing division by 0 in modbulkmicro3_point.f90

jchylik:
Fixing the possible division by 0 in subroutine `sb_evmelt3` in  modbulkmicro3_point.f90, since the calling subroutine (`evapmelting3`) checks only for mass concentration but not for number concentration. Crash reported by @YPChu0723   in test case rf20_composite_setup

### DIFF
--- a/src/modbulkmicro3_point.f90
+++ b/src/modbulkmicro3_point.f90
@@ -2968,7 +2968,7 @@ subroutine sb_evmelt3(avent0,avent1,bvent0,bvent1,x_bmin,n_e,n_ep,n_em &
   g_me= - k_melt*(ktdtodv*(tmp0-T_3)+dvleorv*(esl/tmp0-eslt3/T_3))
 
   ! calculating real mean particle mass
-  x_er = q_e/n_e
+  x_er  = q_e/(n_e+eps0)
 
   ! calculating N_re Reynolds number
   nrex= D_e*v_e/nu_a


YPChu0723 (approved):
The core issue of this commit is resolved. However, I found two new conflicts when running with the current dev version:
1. ltestbed and lforce_user conflict: These two flags are on at the same time, which is causing an issue if we don't compile moduser.f90.

2. lsampcl conflict: When lsampcl is true, it causes modsampling to be confused about which index to use for qr (bulkmicro uses 'iqr' while bulkmicro3 uses 'iq_hr').

jchylik:
Thanks for the update. As for the issues:

> ltestbed and lforce_user conflict: These two flags are on at the same time, which is causing an issue if we don't compile moduser.f90.

That seems to be some outdated call in modtestbed, it does not need lforce_user
I will update modtestbed soon 

> lsampcl conflict: When lsampcl is true, it causes modsampling to be confused about which index to use for qr (bulkmicro uses 'iqr' while bulkmicro3 uses 'iq_hr').

Modsampling in dev branch is currently missing most of the outputs for the case `imicro==imicro_bulk3`. 
It would be better if I update this together with other bulk microphysics statistics.

YPChu0723:
That sounds good!

fjansson:
Sounds good from my side too. Meanwhile I can merge this PR. I wonder though, are there any other places where Nr=0 will lead to problems? I saw that in the warm microphysics, this calculation used to be protected by checks on both qr and Nr, but not anymore - it checks only qr now (cc @CasparJungbacker).

@jchylik if you add stats, Caspar recently re-organized the warm microphysics to add aerosol processes. As a result the tendency statistics could be simplified.

I have added cases for iq_hr in a few places now, where general statistics need a rain field. It's probably not consistent

PR #209, opened 2025-08-21
Fixes for uninitialized memory and switch to own RRTMG_LW

### LINKED ISSUE #19: index out of range in foggy conditions
In rrtmg_lw_rtrnmr.f90, subroutine rtrnmr, there are several suspect-looking accesses to `cldfrac(lev-1)` in the [loop over lev](https://github.com/AER-RC/RRTMG_LW/blob/958edb1ad1618d76d0fcc5a9dc2e3c2f791a54dd/src/rrtmg_lw_rtrnmr.f90#L430
) ending at lev=1. I'm hitting [this one](https://github.com/AER-RC/RRTMG_LW/blob/958edb1ad1618d76d0fcc5a9dc2e3c2f791a54dd/src/rrtmg_lw_rtrnmr.f90#L499) in a DALES run which happens to have cloud in the lowest model layer. A debug build stops with the message
```
At line 499 of file /home/jansson/dales-tester/dales/external/RRTMG/RRTMG_LW/src/rrtmg_lw_rtrnmr.f90
Fortran runtime error: Index '0' of dimension 1 of array 'cldfrac' below lower bound of 1
```

It's of course possible we're doing something wrong on the DALES side. But looking at the RRTMG_LW code, it seems if the lowest level is cloudy there is an access out of bounds.

fjansson:
Various fixes for uninitialized memory

* switch to own fork of RRTMG_LW  until this issue is resolved https://github.com/AER-RC/RRTMG_LW/issues/19
   - fixes nan results from radiation with debug build, and out-of-bounds access in foggy conditions
* fix uninitialized parameters for modscalarpulse - no harm done if the module is not active, but trips the debug build
* modlsm: run the surface scheme for all tiles even when their fraction is 0
* modlsm: move ustar MPI exchange out of 2D loop


fjansson:
wrong branch, this was meant for dev. closing.

PR #210, opened 2025-08-22
Fixes for uninitialized memory and switch to own RRTMG_LW

fjansson:
Various fixes for uninitialized memory

* switch to own fork of RRTMG_LW  until this issue is resolved https://github.com/AER-RC/RRTMG_LW/issues/19
   - fixes nan results from radiation with debug build, and out-of-bounds access in foggy conditions
* fix uninitialized parameters for modscalarpulse - no harm done if the module is not active, but trips the debug build
* modlsm: run the surface scheme for all tiles even when their fraction is 0
* modlsm: move ustar MPI exchange out of 2D loop

### DIFF
--- a/src/modlsm.f90
+++ b/src/modlsm.f90
@@ -72,12 +72,11 @@ subroutine lsm
     call calc_hydraulic_properties
     ! Calculate tendency due to root water extraction
     call calc_root_water_extraction
-    ! Solve diffusion equation:
-    call integrate_theta_soil
 
     ! Update liquid water reservoir
     call calc_liquid_reservoir
-
+    ! Solve diffusion equation:
+    call integrate_theta_soil
 end subroutine lsm
 
 !
--- a/src/modlsm.f90
+++ b/src/modlsm.f90
@@ -1266,6 +1266,7 @@ subroutine integrate_theta_soil
     use modglobal, only : rk3step, rdt, i1, j1, rhow, rlv
     use modsurfdata, only : phiw, phiwm, lambdash, gammash
     use modmpi, only : myidx, myidy
+    use modchecksim, only : check_array
     implicit none
     integer :: i, j, k
     real :: tend, rk3coef, flux_top, fac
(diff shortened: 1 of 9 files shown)


src/modlsm.f90
```diff
@@ -740,7 +739,7 @@ subroutine calc_obuk_ustar_ra(tile)
 
     do j=2,j1
         do i=2,i1
-            if (tile%frac(i,j) > 0) then
+            !if (tile%frac(i,j) > 0) then
```
CasparJungbacker, line 742: Any reason to keep these commented out instead of just removing them? :-)

src/modscalarpulse.f90
```diff
@@ -30,8 +30,12 @@ module modscalarpulse
 ! pulse governing variables (from namelist)
   logical :: lscalarpulse     = .false.
   logical :: lcpmip           = .false.
-  integer(kind=longint) :: timepulse
-  real    :: amppulse, kpulse, zminpulse, zmaxpulse, radius
+  integer(kind=longint) :: timepulse = 3600
+  real    :: amppulse  = 0
```
CasparJungbacker, line 34: Should these be `real(field_r)`?

CasparJungbacker (approved):
What happens exactly when the user sets `lbaseexner=.true.`? Is the exner function then computed just once during startup, and not updated during the simulation?

fjansson:
> What happens exactly when the user sets `lbaseexner=.true.`? Is the exner function then computed just once during startup, and not updated during the simulation?

yes, that's the idea. And they are computed using the *base profiles*. This is somewhat experimental. The reason for doing it is if a pre-processing script needs to convert T to thl, it should use the same exner function as dales to be consistent

src/modlsm.f90
fjansson: I'm not sure yet :) Keeping the if would save some calculations for points where a tile has zero fraction. But I had issues with uninitialized values. In particular 0*nan is nan, so even tiles with fraction=0 should be initialized. And for testing and statistics it might be nicer to calculate full tiles. I think we can do without the ifs.

src/modscalarpulse.f90
fjansson: I didn't think that far. If it fits the calculation that's done with them? Probably yes, since they end up in the prognostic fields.

PR #211, opened 2025-09-02
Refactor transposes

CasparJungbacker:
Moved the transpose operations for the Poisson solver to a separate module to avoid duplicate code. Also wrapped the operations in a derived type for easy implementation of other backends in the future (e.g. cuDecomp).


CasparJungbacker:
Looks like cases with more than 1 halo cell crash. On it...

fjansson (approved):

PR #221, opened 2025-10-22
modstat_nc: add optional zstandard support

fjansson:
new namelist variable NAMNETCDFSTATS/zstandard, default = 0 meaning off. If zstandard is non-zero, enable zstandard compression for netcdf. If zstandard is set, it takes precedence over deflate, which is enabled by default.

Requirements:
* netcdf4 >~ 4.9.2 with zstandard support
* netcdf-fortran >= 4.6.2
* netcdf-fortran installed with the "plugin directory" set.
* DALES compiled with the cmake flag -DENABLE_ZSTANDARD=ON
* saving to the netcdf4 format (lclassic = .false.),

With the cmake flag set, netcdf-fortran >= 4.6.2
is required at compile time. If zstandard is set in the namelist, the other conditions are required,
otherwise a run-time error will occur.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,9 +108,9 @@ if( ENABLE_RRTMG )
   add_compile_definitions( USE_RRTMG )
 endif()
 
-if( ENABLE_ZSTANDARD )
-  add_compile_definitions( NC_ZSTANDARD )
-endif()
+# Zstandard compression
+include( dales_check_zstandard )
+dales_check_zstandard()
 
 # GPU stuff
 
@@ -137,4 +137,5 @@ if( ENABLE_ACC OR ENABLE_NVTX )
 endif()
 
 include( dales_compile_options )
+
 add_subdirectory( src )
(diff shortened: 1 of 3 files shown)


CasparJungbacker:
What happens if the plugin directory for netcdf-fortran isn't set? Does that lead to a compilation error?

fjansson:
My understanding is this:
* without ` -DENABLE_ZSTANDARD=ON` on the cmake command line, everything works as before, no special requirements on netcdf-fortran 

* with ` -DENABLE_ZSTANDARD=ON`, netcdf-fortran >= 4.6.2 is needed to compile (because then it contains the zstandard function). It doesn't matter how it's been compiled or if zstandard will actually work.

* when DALES has been compiled with ` -DENABLE_ZSTANDARD=ON`, and is run with the `zstandard` option set to non-zero in the namelist, the zstandard function is called. This requires that netcdf-fortran has been compiled correctly with the plugin directory set to the same path as for the netcdf-c library, and that netcdf-c has been compiled with zstandard support. Otherwise it gives a runtime error.

PR #223, opened 2025-10-30
modforces: account for gravity on precipitation with bulkmicro3

fjansson:


jchylik:
Yes, that is correct. the same formula in modforces was used in MOSAiC daily simulations.

CasparJungbacker (approved):

PR #228, opened 2025-11-13
Port LSM on GPU with OpenACC

dindon-sournois:
Work in progress


dindon-sournois:
Few notes:
- this was tested on 01_32x32_half_forest_half_grass, which is quite a small test case for GPU
- many timers were added, I don't think we need them all but I'd like to keep them for now, while I investigate performance on a larger test case
- some string comparisons were replaced by a integer comparisons as GPU doesn't support strings and it might also be faster

fjansson (approved):

PR #229, opened 2025-11-28
Move Poisson solver namelist and data

CasparJungbacker:
Small cleanup

- Moved data for the Poisson solver from `modglobal` to a new module `modpois_data`
- Added subroutine for reading solver namelist, called from `read_namelists`


CasparJungbacker:
nvidia pipelines failing because there's no space on the runners to install the full HPC sdk...

fjansson (approved):

PR #231, opened 2025-12-02
Modlogging: add better logging support

Xetalim:
Large diff, sorry about that. Please test around a bit, and see if it fails anywhere, and look through your own favorite sections of code if you see anything wrong.
In summary:

- `stop` has been replaced with `call finish(routine, error_message)` everywhere, which then prints a backtrace. I recommend using `--output-filename <filename>` to separate the logs on large clusters. See: [https://www.open-mpi.org/doc/v4.1/man1/mpirun.1.php](https://www.open-mpi.org/doc/v4.1/man1/mpirun.1.php).
- all namelists are dumped to a file named "namelist_used.txt" instead of stderr/stdout
- all profile information is dumped to a file named "profiles_used.txt" instead of stderr/stdout.


cmake/dales_add_fortran_support.cmake
```diff
@@ -0,0 +1,30 @@
...
+    libfortran-support_ep
+    GIT_REPOSITORY https://gitlab.dkrz.de/icon-libraries/libfortran-support.git
+    GIT_TAG 2.2.0
+    CMAKE_ARGS
+        -DCMAKE_INSTALL_PREFIX=${EXTERNAL_INSTALL_DIR}
+        -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
+        -DBUILD_SHARED_LIBS=OFF
+        -DBUILD_TESTING=OFF
+        -DFS_ENABLE_BACKTRACE_TEST=ON
+        -DFS_ENABLE_OPENACC=ENABLE_OPENACC
```
CasparJungbacker: `ENABLE_OPENACC` should be `${ENABLE_ACC}`. Right now, Intel pipeline is failing because `fortran-support` enables OpenACC by default which isn't available from the Intel compiler.

CasparJungbacker (changes requested):

cmake/dales_add_fortran_support.cmake
Xetalim: Good catch, thanks! Would the same be true for `-DFS_ENABLE_SINGLE_PRECISION=FP32_FIELDS`? I'm assuming that is the case and have edited both switches.

CasparJungbacker: Hmm, I would think so. I don't know if it then compiles the library in both double and single precision, or only in single precision?

CasparJungbacker (approved):

fjansson (approved):

PR #234, opened 2025-12-03
OpenACC aerosols

CasparJungbacker:
Adds OpenACC support to the M7 aerosol module.


CasparJungbacker:
One note: I didn't make the kernels asynchronous yet, because the bulkmicro kernels also aren't asynchronous

fjansson (approved):

PR #235, opened 2025-12-05
Clean up surface namelist

CasparJungbacker:
Episode 2 of The Big Cleanup:

- Added `surface_read_namelist`, called from `modnamelist` to make all surface options available from the start.
- Deleted surface options from namelist `PHYSICS`, instead they should be provided in `NAMSURFACE`, where they were also read from.
- Fixed all cases for the previous change.


fjansson (approved):
Nice! The duplication has bothered me, but I never realized it was so many variables!

PR #247, opened 2026-01-23
Add a new property (`laero`) to the tracer manager to indicate aerosol scalar tracers.

adoyenne:
Changes include:

- `modtracer_type.f90`: added a logical property `laero` in the tracer type.  
- `modtracers.f90`: updated routines (`add_tracer`, `allocate_tracers`, `tracer_props_from_ascii`, `tracer_props_from_netcdf`, and print routines) to handle the new `laero` property.  
- The **`is_aerosol` column is now required in the tracer setup input file** (`tracerdata`).  
- Existing tracers remain backward-compatible; only tracers marked as aerosol need `is_aerosol = .true.`.  

This ensures that aerosol scalar tracers (e.g., PM2.5) are managed explicitly throughout the model and properly initialized, allocated, and printed.

### DIFF
--- a/src/modtracer_type.f90
+++ b/src/modtracer_type.f90
@@ -15,6 +15,7 @@ module modtracer_type
     logical           :: ldep = .false.     !< Boolean if tracer is deposited
     logical           :: lags = .false.     !< Boolean if in A-gs
     logical           :: lmicro = .false.   !< Boolean if in cloud microphysics
+    logical           :: laero = .false.    !< Boolean if in aerosol
     real(field_r)     :: wsvsurf = 0        !< Kinematic surface flux (- m/s)
     logical           :: lnudge = .false.   !< Boolean if tracer is nudged
   contains
@@ -36,6 +37,7 @@ contains
     write(*,*) "  ldep       : ", self%ldep
     write(*,*) "  lags       : ", self%lags
     write(*,*) "  lmicro     : ", self%lmicro
+    write(*,*) "  laero      : ", self%laero
     write(*,*) "  wsvsurf    : ", self%wsvsurf
   end subroutine tracer_print_properties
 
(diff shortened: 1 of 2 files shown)


CasparJungbacker:
Did you add this with the M7 aerosol scheme (`modaerosol.f90`) in mind?

adoyenne:
> Did you add this with the M7 aerosol scheme (`modaerosol.f90`) in mind?

No, the current implementation I added independently for a scalar aerosol tracer to allow easy distiguish chem/aerosol tracers for separate treatment in the dry deposition and emission modules (as the units differ from chemical tracers) and does not modify the M7 aerosol scheme (modaerosol.f90). So, now it is not interact directly with modaerosol.f90, but could potentially be integrated with the M7 framework in the future via tracer manager, if needed

CasparJungbacker:
Aha, got it. Is there a way you could use the existing `unit` attribute of the tracers instead? I'd prefer to keep the `laero` attribute reserved for stuff that interacts with the M7 scheme, to prevent confusion later on.

adoyenne:
> Aha, got it. Is there a way you could use the existing `unit` attribute of the tracers instead? I'd prefer to keep the `laero` attribute reserved for stuff that interacts with the M7 scheme, to prevent confusion later on.

Yes, in principle we can rely only on the unit attribute and skipping this update, if it is a bit difficult now to connect the separation logic to aerosols via the tracer manager.
I was just thinking it might be more practical and consistent to separate tracers via a dedicated flag in tracerdata, but I’m fine to align with keeping laero reserved for M7-related tracers.

CasparJungbacker:
> I’m fine to align with keeping laero reserved for M7-related tracers.

Ok, let's do that then! I see you submitted another PR (#250), does that fix your issue with the units?

adoyenne:
> > I’m fine to align with keeping laero reserved for M7-related tracers.
> 
> Ok, let's do that then! I see you submitted another PR (#250), does that fix your issue with the units?

Yes, for now, I switched to the simplified approach by just using if trim(tracer_prop(l)%unit) == 'kg m-3' to inform the model that this scalar tracer should be treated as aerosol

CasparJungbacker:
Cool! Closing this PR then...

PR #248, opened 2026-01-23
moddatetime: add starthour and rtimee to support a proper datetime handling

adoyenne:
Description:
Update for the moddatetime module to include support for starthour in the namelist and add the rtimee to datetime calculation. 

Key changes include:

Added starthour variable to the DATETIME namelist.

Updated datetime initialization to correctly account for starthour when computing the hour and day, including timezone adjustments. rtimee is added to the date calculation.

Added handling notes for rtimee at warmstart. Currently, rtimee is always 0 in this module because it is read from restart files after moddatetime is called (TODO for proper future implementation). When performing a warmstart, the date in the namelist should be set manually to the warmstart time to ensure consistency.

This ensures that simulations starting at non-zero hours are correctly initialized and lays the groundwork for accurate handling of warmstart/restart times.

### DIFF
--- a/src/moddatetime.f90
+++ b/src/moddatetime.f90
@@ -36,6 +36,7 @@ module moddatetime
   integer :: startyear  = 0, &
              startmonth = 0, &
              startday   = 0, &
+             starthour  = 0, &
              timezone   = 0  
 
   integer               :: jday
@@ -47,17 +48,17 @@ module moddatetime
     ! Using the namelist, construct datetime relevant variables
 
     use modmpi,    only : myid, comm3d, D_MPI_BCAST
-    use modglobal, only : ifnamopt, fname_options, xtime
+    use modglobal, only : ifnamopt, fname_options, xtime, rtimee
     use fortran_support, only : nnml_output
    
     implicit none
     character(len=*), parameter :: routine = modname//'/initdatetime'
    
     ! Auxiliary variables
-    integer :: ierr
+    integer :: ierr, itimee
 
     ! Read & broadcast namelist DATETIME -----------------------------------
-    namelist/NAMDATETIME/ l_datetime, startyear, startmonth, startday, timezone
+    namelist/NAMDATETIME/ l_datetime, startyear, startmonth, startday, starthour, timezone
 
     if (myid == 0) then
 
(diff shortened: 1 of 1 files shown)


jchylik:
>  Currently, rtimee is always 0 in this module because it is read from restart files after moddatetime is called (TODO for proper future implementation)

Are you sure?

At the end of **readinitfiles** in modstartup, variable `rtimee` is calculated from variable `timee`:
[modstartup.f90#L1164](https://github.com/dalesteam/dales/blob/dev/src/modstartup.f90#L1164)

`timee` is read from restart files:
[modstartup.f90#L1229](https://github.com/dalesteam/dales/blob/dev/src/modstartup.f90#L1229)

`timee` is written to restart files:
[modstartup.f90#L1404](https://github.com/dalesteam/dales/blob/dev/src/modstartup.f90#L1404)

adoyenne:
> > Currently, rtimee is always 0 in this module because it is read from restart files after moddatetime is called (TODO for proper future implementation)
> 
> Are you sure?
> 
> At the end of **readinitfiles** in modstartup, variable `rtimee` is calculated from variable `timee`: [modstartup.f90#L1164](https://github.com/dalesteam/dales/blob/dev/src/modstartup.f90#L1164)
> 
> `timee` is read from restart files: [modstartup.f90#L1229](https://github.com/dalesteam/dales/blob/dev/src/modstartup.f90#L1229)
> 
> `timee` is written to restart files: [modstartup.f90#L1404](https://github.com/dalesteam/dales/blob/dev/src/modstartup.f90#L1404)

Yes, I checked that. In initdatetime, rtimee is always 0, regardless of whether it’s a cold or warm start. This means that if input data rely on datetime, they will always be read for the cold time, not the warm start time. Currently, the only workaround is to manually adjust the datetime in namoptions to match the warm start time, which is inconvenient.
I think the proper fix would be to move the call initdatetime to after rtimee has been read from the restart file. That way, initdatetime will use the correct rtimee.

jchylik:
> In initdatetime, rtimee is always 0

Oh, I see, in `initdatetime` it is 0, and in subroutine `datetime` it is correct.

> I think the proper fix would be to move the call initdatetime to after rtimee has been read from the restart file. 

I would be a solution, but before the array `datex` is used, it is neverthess updated by the call of `datetime`.

fjansson:
One thing I'm hesitant about is that there is already xday and xtime in the DOMAIN namelist, also for setting the starting time. This feels like a duplication.

adoyenne:
> One thing I'm hesitant about is that there is already xday and xtime in the DOMAIN namelist, also for setting the starting time. This feels like a duplication.

well, the reason I added starthour is that xtime should always reflect the cold start time. However, in initdatetime, rtimee is always 0 (it is read from restarts after the call to initdatetime), even for a warm start.

Thus, starthour is needed to read emissions correctly for warm starts, accounting for the warm start time. If there were a way to put a correct rtimee to initdatetime (instead of it always being 0), we could likely skip adding starthour. Without this, we cannot read the emissions at the correct times for warm starts..

fjansson:
After looking a bit more, I like the date handler. Would it be possible to fix this by changing the initialization order, so that the date dandler would know the current rtimee at a restart? I'd rather do that than add a namelist option.

adoyenne:
> After looking a bit more, I like the date handler. Would it be possible to fix this by changing the initialization order, so that the date dandler would know the current rtimee at a restart? I'd rather do that than add a namelist option.

I tried changing the call positions in startup, but it always resulted in a crash due to dependency/order issues, so it’s not that straightforward. Perhaps you could check whether it’s possible to place the call to initdatetime after reading rtimee from the restart files?

PR #251, opened 2026-02-03
Update RTE-RRTMGP to version 1.9.2

CasparJungbacker:
Primarily to fix compilation issue on LUMI. Also took the opportunity to clean up the CMakeLists a bit. RRTMGP is now handled in a similar fashion as `fortran-support`: we build it as a separate library and then link it to DALES.


fjansson:
I tried this branch on the deep-botany sample case with `iradiation` changed to 5, this gave a thermo crash on start. This was a small case on a single CPU core, gfortran, single precision. With `iradiation = 4` it started normally.
Btw, do we still use RRTMGP data files from external/rrtmgp-data? Thats' what I did for this test.

CasparJungbacker:
@fjansson Did you compile RRTMGP in single precision as well?

fjansson:
> @fjansson Did you compile RRTMGP in single precision as well?

No (as in I didn't change any settings for that, just the normal cmake flags to compile DALES itself in sp).
The crash happens also when I compile DALES in the default dp mode.

fjansson (approved):

fjansson:
Approved PR, it turns out the crash mentioned above happens on only one computer so far (gfortran 15.2.1), and there it happens for both dev and this branch.

PR #255, opened 2026-02-07
Refactor sprayer, couple to aerosols

CasparJungbacker:
- Added option to spray monodisperse aerosols
- Added a separate module for the lateral sponge layer routines
- Refactoring, formatting, `field_r`, etc...


fjansson (approved):
possibly the latsponge routine needs to be more careful in the corners, to not nudge twice.

PR #257, opened 2026-02-12
modchecksim: fix automatic exit of simulation if lstop=.true.

Xetalim:
This fixes a bug that was probably introduced in [https://github.com/dalesteam/dales/commit/4ceba11](https://github.com/dalesteam/dales/commit/4ceba11).
Currently if lstop=true, the simulation crashes immediately. If the value is acceptable, we should cycle, but now the cycle is not hit if a threshold is defined.


CasparJungbacker:
LGTM

PR #261, opened 2026-02-17
PR: add urban tile to the LSM

Xetalim:
Still needs the occasional tests here and there, but there's plenty of warnings so you don't accidentally make big mistakes...
Energy balance has been checked to be closed, and it seems to run quite well!

Also adds:
- Reading of arbitrary netcdf fields with read_nc_field
- allows program exit if value in field is outside bounds
- Fixs a bug in init_from_netcdf (it read in height->zf from 1 to k1, even though it should read from 1 to kmax, and fill in k1 with some value calculated in modglobal)


jchylik:
Is it related to the UDales project? [^1]

[^1]: DALES for urban airflow https://github.com/uDALES/u-dales

Xetalim:
No, but it is related to the UrbanAIR project 😀 
UDales has very fancy 3d radiative calculations, and tree representations at resolutions of 1 or 2 meters, while this urban tile is more meant for lower resolutions of between 20 and 100 meters. It calculates an urban energy balance on an urban canyon assumption. It is adapted from the PALM SLUrb (single layer urban canopy), which is another LES out there.

fjansson (approved):

PR #263, opened 2026-02-23
Small fixes in thermodynamics

CasparJungbacker:
Fixes an issue where `interp_w` went out of the range (0,1), giving too high values for the saturation humidity.


fjansson:
There are some old variable names left in the ACC statement on line 883. Looks good otherwise!

fjansson (approved):

PR #267, opened 2026-03-02
HYPRE: add hypre initialize function (to fix crash)

StevenvdLinden:
Added explicit call to the HYPRE initialize function. Without this (unexpected) crashes may occur.

### DIFF
--- a/src/modhypre.f90
+++ b/src/modhypre.f90
@@ -54,7 +54,7 @@ save
 ! for fortran
 interface
   subroutine HYPRE_Initialize( ierr) &
-    bind(c, name="hypre_initialize_")
+    bind(c, name="hypre_init_")
     use iso_c_binding
     implicit none
     integer(c_int) ierr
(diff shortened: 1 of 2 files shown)


StevenvdLinden:
Fails automated hypre test in which libhypre-dev is automatically installed. In my testbuilds, I built latest Hypre release locally on machine

fjansson:
Possibly `HYPRE_Initialize()` is only present in the new version. There is a [`HYPRE_Init()`](https://hypre.readthedocs.io/en/latest/api-utilities.html#c.HYPRE_Init) "for backwards compatibility". Perhaps that one is available in both the old and new versions? I'd rather not make DALES incompatible with HYPRE 2.x just yet, since that's what you get in fairly recent ubuntu for example. 
We could do some #ifdef logic, but that feels like a lot of variants to maintain for something we don't use a lot.

fjansson:
Replacing ` HYPRE_Initialize` by ` HYPRE_Init` makes it compile on Ubuntu 22 with HYPRE 2.28. @StevenvdLinden can you do that change and try on your machine with the new hypre?

fjansson (approved):

PR #272, opened 2026-03-16
New NetCDF interface + support for parallel NetCDF

CasparJungbacker:
- Added `netcdf_file_t` which is an abstract derived type containing all relevent NetCDF file info. Essentially, this wraps the boiler plate code that was needed for every statistical module (`ncname`, `vars`, calls to `ncinfo`, `define_nc` etc.)
- Added implementations of `netcdf_file_t` for time series, profile, cross section and field dump files. These contain a buffer with the correct dimensions that replace the old `vars` arrays. The buffer can be written to by setting up a pointer using the `get_pointer` method of the derived types.
- Added `modstat_nc_files.f90` (name due to change, don't like this one), which keeps track of all output files and does the timekeeping.
- Updates to `modstat_nc.f90` to support opening NetCDF files in parallel.
- Added `modcloudstat.f90`, which is a re-implementation of some statistical quantities of `modcape.f90`. This module makes use of the new NetCDF interface.
- Added CMake script to automatically deduce if the selected NetCDF library has parallel capabilities. If not, opening files in parallel is disabled in DALES.

TODO:
- [x] Re-opening an existing file in parallel doesn't work yet. NetCDF crashes with an HDF error on the first write.
- [x] Handle parallel output of vertical cross sections. Right now, if parallel output is enabled, cross section files are opened using the 3D communicator while only a subset of the MPI ranks will actually have data to write. This shouldn't be a problem if independent access is enabled for the variable to write (the default), but one needs to make sure that the correct subset of MPI ranks makes the call to `writestat_nc`.
- [x] Add André's improvements from #265.
- [ ] Update all statistical modules to make use of the new interface. (doesn't have to be done immediately, but I'd say let's do at least timestat + genstat + cross + fielddump)
  - [ ] genstat
  - [ ] timestat
  - [x] fielddump
  - [x] cross sections
  - [x] agscross
  - [x] depcross
  - [ ] lsmstat
  - [ ] samptend
  - [ ] microphysics stats
  - [x] cape
  - [ ] radstat
  - [x] radfield
  - [ ] msebudg
  - [ ] budgetstat
  - [ ] varbudget
  - [ ] leibniztend?
  - [ ] quadrant
- [x] Add OpenACC directives
- [x] Add clean-up routine for `modstat_nc_files`
- [ ] Benchmarking
- [ ] Re-add `loutdir` functionality


CasparJungbacker:
Setting `deflate=0` in `NAMNETCDFSTATS` fixes the HDF crash. Now, it's crashing with the following error:

```
FINISH PE:     0 modstat:: NetCDF: Attempt to extend dataset during NC_INDEPENDENT I/O operation
```

which I already suspected from looking at the PALM NetCDF code. I.e., we cannot use independent variable access mode on variables with unlimited dimensions. I see two options here:

1) Fix the length of the time axis when defining the files. For restarting, we'd have to find some way to extend the time axis.
2) Switch to collective access for all variables. This would be fine for xy cross sections and field dumps, but requires some MPI voodoo for vertical cross sections, since those are only written by a subset of the ranks.

src/modradrrtmg.f90
Xetalim, line 1: this is an erroneous merge, emis is already set in line 698

PR #275, opened 2026-03-19
add albedo variable to copyin

fjansson:


Xetalim:
I have no device to test openACC things on(and little openACC knowledge), so I'm not sure which pragmas to add where, but do we need also an ACC copyin here for albedo?

https://github.com/dalesteam/dales/blob/dev/src/modlsm.f90#L1655

CasparJungbacker:
@Xetalim there is an `!$acc enter data copyin(tile(ilu)%albedo)` on line 1742, wouldn't that be enough? Or is the array `albedo` also needed when using the LSM?

Xetalim:
The albedo array is also needed, as the LSM updates the albedo array from the LSM tiles (that may have time-varying albedos):
                    albedo(i,j) = albedo(i,j) + tile(ilu)%frac(i,j) * tile(ilu)%albedo(i,j)

Xetalim:
I added this to #276, so I'm closing this PR

PR #276, opened 2026-03-27
Fix some bugs in modslurb, and allow varying emissivities.

Xetalim:
I'll probably break some openACC compilation again, let's see!


Xetalim:
SP compilation should be fixed now, also did some more tweaks, should work some better now.

src/modslurb/modslurb.f90
```diff
@@ -291,8 +291,8 @@ SUBROUTINE slurb_update_external_vars
 
             du = 0.5*(u0(i,j,1) + u0(i+1,j,1)) + cu
             dv = 0.5*(v0(i,j,1) + v0(i,j+1,1)) + cv
-            slurb_tile%uv_abs1(i,j) = sqrt(du**2 + dv**2)
-            ! slurb_tile%uv_abs1(i,j) = max(0.1, sqrt(du**2 + dv**2)) DALES VERSION
+            ! slurb_tile%uv_abs1(i,j) = sqrt(du**2 + dv**2)
+            slurb_tile%uv_abs1(i,j) = max(0.1, sqrt(du**2 + dv**2))! DALES VERSION
```
fjansson, line 295: there is a variable for the 0.1 constant called min_horv. It's possibly not used completely consistently...

fjansson (approved):

src/modslurb/modslurb.f90
Xetalim: I see! It's also not used in the LSM..

Xetalim:
I went ahead and added emissivity and albedo from #275.

PR #278, opened 2026-04-01
Modnudge: fix wrong read for w, only allocate to kmax, we didnt use k1

Xetalim:


Xetalim:
This was accidentally included in #276, closing.

PR #279, opened 2026-04-02
modslurb: workaround, avoid initialization with nan on Fujitsu compiler

fjansson:
...because those seem to crash the compiler.


Xetalim (commented):
LGTM

PR #280, opened 2026-04-04
fixes for Fujitsu compiler

fjansson:
small fixes for the Fujitsu compiler to work, and fix some uninitialized variables found by the Fujitsu debug build.

### DIFF
--- a/src/modradrrtmg.f90
+++ b/src/modradrrtmg.f90
@@ -156,7 +156,7 @@ contains
       end if
       call readTraceProfs
 
-      call message(routine, 'Trace gas profile have been read')
+      call message(routine, 'Trace gas profile have been read',all_print=.false.)
       isReadTraceProfiles = .true.
     end if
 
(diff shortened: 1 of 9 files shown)


fjansson:
Compilation and running on Fugaku also requires the more intrusive changes in branch [dev-fujitsu-hacks](https://github.com/dalesteam/dales/commits/dev-fujitsu-hacks/)

src/modradrrtmg.f90
```diff
@@ -156,7 +156,7 @@ subroutine radrrtmg
       end if
       call readTraceProfs
 
-      call message(routine, 'Trace gas profile have been read')
+      if (myid == 0) call message(routine, 'Trace gas profiles have been read')
```
CasparJungbacker: `message` also has an optional argument `all_print`, which can be set to `.false.` to limit printing to the root process.

fjansson: I've rebased this PR's branch onto the current dev, and changed it to use the all_print argument.

Xetalim (commented):
Looks good to me!

CasparJungbacker (approved):

PR #287, opened 2026-04-16
Interpolate thl and qt to half levels in dry cases

CasparJungbacker:
Interpolation was only done in cases where `lmoist = .true.`, causing a crash in `modforces` because the buoyancy array was filled with garbage.


StevenvdLinden (approved):
Looks good

fjansson (approved):

PR #294, opened 2026-06-03
bugfix: add missing lmoist guard on qtsurf-calls 

StevenvdLinden:
Safety guard to prevent unexpected qts/qskin behavior and thus influence on surface temperature when lmoist=.false.


CasparJungbacker:
@StevenvdLinden Circular dependency 🙃 I think it's because `modsurface` needs `modthermodynamics`, which needs `modibm`, which needs `modsurface` again. I think the solution here is to move thermo switches to a public module.

StevenvdLinden:
woops.. that is a silly oversight. I'll think about moving these switches to a public module indeed

PR #297, opened 2026-06-16
modlogging: switch to simple print for logging calls on Fujitsu

fjansson:
...to avoid segfault in Fortran support library


fjansson:
This workaround is no longer needed, after passing the compiler flags to fortran-support to enable allocation on assignment. Closing.

PR #302, opened 2026-06-26
Parallel netcdf: colstat, virtualmeasurement and warmstart fixes

Xetalim:
To be merged in parallel netcdf. Contains some cleanup for modstat_nc, and (optional) chunking behavior.

### DIFF
--- a/src/modcrosssection.f90
+++ b/src/modcrosssection.f90
@@ -81,12 +81,12 @@ contains
     namelist /NAMCROSSSECTION/ lcross, dtav, crossheight, crossplane, &
                                crossortho, lxy, lxz, lyz
 
-    crossheight(1)=2
-    crossheight(2:100)=-999
-    crossplane(1)=2
-    crossplane(2:100)=-999
-    crossortho(1)=2
-    crossortho(2:100)=-999
+    crossheight(1) = 2
+    crossheight(2:100) = -999
+    crossplane(1) = 2
+    crossplane(2:100) = -999
+    crossortho(1) = 2
+    crossortho(2:100) = -999
 
     dtav = dtav_glob
 
--- a/src/modcrosssection.f90
+++ b/src/modcrosssection.f90
@@ -81,9 +81,12 @@ contains
     namelist /NAMCROSSSECTION/ lcross, dtav, crossheight, crossplane, &
                                crossortho, lxy, lxz, lyz
 
-    crossheight(1:100) = -999
-    crossplane(1:100) = -999
-    crossortho(1:100) = -999
+    crossheight(1)=2
+    crossheight(2:100)=-999
+    crossplane(1)=2
+    crossplane(2:100)=-999
+    crossortho(1)=2
+    crossortho(2:100)=-999
 
     dtav = dtav_glob
 
--- a/src/modcrosssection.f90
+++ b/src/modcrosssection.f90
@@ -81,12 +81,9 @@ contains
     namelist /NAMCROSSSECTION/ lcross, dtav, crossheight, crossplane, &
                                crossortho, lxy, lxz, lyz
 
-    crossheight(1) = 2
-    crossheight(2:100) = -999
-    crossplane(1) = 2
-    crossplane(2:100) = -999
-    crossortho(1) = 2
-    crossortho(2:100) = -999
+    crossheight(1:100) = -999
+    crossplane(1:100) = -999
+    crossortho(1:100) = -999
 
     dtav = dtav_glob
 
--- a/src/modcrosssection.f90
+++ b/src/modcrosssection.f90
@@ -22,7 +22,7 @@ module modcrosssection
   use modlogging,   only: nnml_output, finish, warning
   use modglobal,         only: longint, kmax, nsv, cu, cv, itot, jtot, imax, &
                                jmax, kmax, i1, j1, ifnamopt, dtav_glob, &
-                               rk3step, checknamelisterror, dx, dy, zf, x0, y0
+                               rk3step, checknamelisterror, dx, dy, zf, x0, y0, output_prefix
   use modtracers,        only: tracer_prop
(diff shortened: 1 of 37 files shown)


Xetalim:
To be continued after the weekend...

Xetalim:
I've fixed the SP compilation, but I don't really like the solution. If any of you have a better idea of how to do this, let me know!

Xetalim:
Last commit should have fixed all SP issues, so should be ready for review :)

src/modcrosssection.f90
```diff
@@ -81,12 +81,12 @@ subroutine crosssection_read_namelist(nml_filename)
     namelist /NAMCROSSSECTION/ lcross, dtav, crossheight, crossplane, &
                                crossortho, lxy, lxz, lyz
 
-    crossheight(1) = 2
-    crossheight(2:100) = -999
-    crossplane(1) = 2
-    crossplane(2:100) = -999
-    crossortho(1) = 2
-    crossortho(2:100) = -999
+    crossheight(1)=2
```
CasparJungbacker: Why *remove* spaces???

src/modlsmdata.f90
```diff
@@ -78,14 +80,15 @@ module modlsmdata
       ! Dynamic tile fraction:
       real, allocatable :: frac(:,:)
       ! Monin-obukhov / surface layer:
-      real, allocatable :: obuk(:,:), ustar(:,:), ra(:,:)
+      real(field_r), allocatable :: obuk(:,:)
+      real, allocatable :: ustar(:,:), ra(:,:)
```
CasparJungbacker, line 84: Can't these be field_r as well?

src/modstat_nc_files.f90
```diff
@@ -124,27 +125,38 @@ end subroutine init_output_files
 
   !> Loop over the files and write those that are due.
   subroutine write_output_files()
-
+    use modtimer, only: timer_tic, timer_toc
```
CasparJungbacker: Please put this at the top of the module with the other use statements :-)

src/modstat_nc_files.f90
```diff
@@ -124,27 +125,38 @@ end subroutine init_output_files
...
     integer :: ifile
-
+    call timer_tic("write_output_files")
     do ifile = 1, nfiles
+      call timer_tic("write_output_files_file_"//trim(file_list(ifile)%file%filename))
```
CasparJungbacker: Ahhh, timers per file. I like it

src/modsurface.f90
```diff
@@ -1743,18 +1743,18 @@ pure function psim(zeta)
...
-    real, intent(in) :: zeta
-    real             :: x
+    real(field_r)             :: psim
+    real(field_r), intent(in) :: zeta
+    real(field_r)             :: x
 
     if(zeta <= 0) then
-      x     = (1. - 16. * zeta) ** (0.25)
-      psim  = 3.14159265 / 2. - 2. * atan(x) + log( (1.+x) ** 2. * (1. + x ** 2.) / 8.)
+      x     = (1._field_r - 16._field_r * zeta) ** (0.25_field_r)
```
CasparJungbacker: Integers don't have to be casted to field_r explicitly, the compiler will take care of this. It's not wrong, but not doint it can keep lines like these a bit shorter and more readable

src/modtimestat.f90
```diff
@@ -740,9 +751,9 @@ subroutine timestat
 
     if (lhetero) then
       do k = 1, kmax
-        u0av_patch = patchsum_1level_field_r(u0(2:i1,2:j1,k)) * (xpatches*ypatches/ijtot)
-        v0av_patch = patchsum_1level_field_r(v0(2:i1,2:j1,k)) * (xpatches*ypatches/ijtot)
-        w0av_patch = patchsum_1level_field_r(w0(2:i1,2:j1,k)) * (xpatches*ypatches/ijtot)
+        u0av_patch = patchsum_1level(u0(2:i1,2:j1,k)) * (xpatches*ypatches/ijtot)
```
CasparJungbacker, line 754: The `lhetero` stuff is already gone in `dev`, so this might lead to merge conflicts.

CasparJungbacker (commented):
I have one general question about the new statistics modules: why do you setup (allocatable) variables for all quantities, instead of using pointers and the `get_pointer` method from the various NetCDF file type wrappers?

src/modlsmdata.f90
Xetalim: I agree, but we might want to separate that out in a different PR? Cleaning up some of the extra field_r stuff

src/modstat_nc_files.f90
Xetalim: I moved them :)

src/modsurface.f90
Xetalim: I changed them back to normal integers, seems like a good idea

src/modtimestat.f90
Xetalim: I think it's better to have the merge conflict later, as it's nice to know that it compiles correctly on this branch, as not doing anything there will lead to SP comp errors

src/modstat_nc_files.f90
Xetalim: I like them as well but I unfortunately had to remove them, because when you use crosssections, it might be that 1 core has more timers than another, which leads to a hang in the timer writes at the end :/

src/modcrosssection.f90
Xetalim: Good point, I had originally removed this initialization towards cross*(1)=2, but that broke the default behaviour, so I brought it back (wrongly). Fixed now!

Xetalim:
I moved towards using the pointer manager instead.  The only possibly nasty thing is that the scalar value calculations have to get a new pointer each time we move to a different column index.

PR #304, opened 2026-07-01
Add qsat to restart files to enable warm starts with deposition

LeonGeers:
Ruud found a problem running warm starts with DALES, specifically when deposition was activated. It turned out that the compensation point concentration was NaN, caused by the fact that qsat was not defined when restarting (it was not included in the warm start initializer files). He's corrected that for DALES v4.4 (Ruisdael branch) and I cherry-picked it for version 5.


fjansson:
Good find, let's fix this.  However, this pull request is massive because it tries to merge into the main branch. Probably it should target dev, with just the one cherry-picked commit.

Second, qsat can be calculated from the prognostic variables, so it might be possible to not store it in the restart file but make sure it's been calculated in the time step before it's required - that's nice also otherwise because then the qsat value used is the most recent one. Not sure if this is easy or possible in this case.

LeonGeers:
Oh, crap. I was under the impression that, since it is a cherry-pick to a sidebranch of dev, it should generate a merge request in dev. F'd it up again, apparently. I'll have a look at it tomorrow

LeonGeers:
qsat is calculated in every time step, a fix that was made some time ago, which renders this merge request redundant.

PR #305, opened 2026-07-03
Fix re-compilation issue with `libfortran-support`

CasparJungbacker:
Prevents CMake from performing the update step for `libfortran-support` upon rebuilds.


Xetalim (approved):
Looks good to me

PR #307, opened 2026-07-06
Some fixes for the LSM and slurb

Xetalim:
the uppercase/lowercase commit contains only uppercase/lowercase changes, so if the diff is large, try viewing somehow the PR without that one :D


fjansson (approved):
Looks good.

There are some capitalized words left, can be fixed later.
IMPLICIT NONE, EXIT, LOGICAL
HUGE, TINY, SQRT, MAX, MIN, LOG

PR #308, opened 2026-07-07
GPU execution guards

CasparJungbacker:
Added errors in modules that haven't been offloaded yet.

### DIFF
--- a/src/moddepcrosssection.f90
+++ b/src/moddepcrosssection.f90
@@ -84,7 +84,9 @@ contains
    end if
 
 #if defined(DALES_GPU)
-    call finish(routine, "depcrosssection not supported on GPU")
+    if (ldepcrosssection) then
+      call finish(routine, "depcrosssection not supported on GPU")
+    end if
 #endif
 
     call D_MPI_BCAST(dtav,             1, 0, comm3d, mpierr)
--- a/src/moddepcrosssection.f90
+++ b/src/moddepcrosssection.f90
@@ -83,6 +83,10 @@ contains
         & and/or land surface model defined"
    end if
 
+#if defined(DALES_GPU)
+    call finish(routine, "depcrosssection not supported on GPU")
+#endif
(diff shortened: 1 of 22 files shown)


src/modslurb/modslurb.f90
Xetalim, line 207: put the guard after here instead of the beginning of initslurb, as initslurb will always be called

src/moddepcrosssection.f90
```diff
@@ -83,6 +83,10 @@ subroutine initdepcrosssection
         & and/or land surface model defined"
    end if
 
+#if defined(DALES_GPU)
+    call finish(routine, "depcrosssection not supported on GPU")
```
Xetalim: should this be called only if ldepcrosssection?

src/moddrydeposition.f90
```diff
@@ -103,6 +105,10 @@ subroutine initdrydep
   call d_mpi_bcast(nh3_avg,              1, 0, comm3d, ierr)    !GT added
   call d_mpi_bcast(so2_avg,              1, 0, comm3d, ierr)    !GT added
 
+#if defined(DALES_GPU)
+  call finish(routine, "deposition is not supported on GPU")
```
Xetalim: same with this one

src/modslurb/modslurb.f90
CasparJungbacker: Ah ok, I thought it was only called if the LSM was enabled

Xetalim (changes requested):
there's a few places where the finish call should be after the early return, instead of before the early return if I'm correct in reading this

CasparJungbacker:
Yep, you were right. Should be fixed now, thanks 👍

Xetalim (approved):
Looks good now!

fjansson (approved):

Xetalim:
should also be merged now

PR #313, opened 2026-07-17
Modstartup: refactor baseprofs and keep exner function constant

Xetalim:
Disable time-varying exner function

You can either choose one of the previous reference state options, or choose to get the reference state by integrating your chosen thl profile assuming a dry atmosphere, or you can put in a var baseprof_T in init nc, which is the real temperature profile which is then hydrostatically integrated. The exner function is then based on the baseprof P. Also possible to supply both baseprof_T and baseprof_P, where one has to correctly and hydrostatically calculate those. Especially useful when using open boundary conditions.

I'm marking this as a draft, just so we can discuss well enough to be sure that this is physically correct. Although I think it's fine, we need to make sure.


fjansson (approved):
I've ran this successfully on deep-botany cases, with ibas_prf=3 (default) and ibas_prf=6 (initial thl profile as base profile).

PR #315, opened 2026-07-19
fix cape thermo hang and the formatting of numbers in checktend

fjansson:
* chektend warnings had a too short buffer, the number exponent wasn't always printed.
* fix hang in cape thermodynamics calculation


Xetalim (approved):

CasparJungbacker (approved):
LGTM

PR #316, opened 2026-07-21
corrected some physical units definitions

LeonGeers:
Some of the physical units defined in the attributes of NetCDF files were not formatted correctly (or at least unconventionally). Processing the output files of DALES using unit conversion routines then fails. Things like `Km` (as a unit!) I changed to `K m`. but worse is `/kg/s` which should be `1/kg/s` and `kg/ms^2` which should be `kg/m/s^2` to not confuse meter second with millisecond.


fjansson (approved):
Looks good.

fjansson:
@LeonGeers Is there a specific standard we should follow for the unit formatting? The change kg/ms^2 -> kg/m/s^2 makes sense for machine reading but isn't obvious to me. I had a quick look at the cf convention but didn't find something.

LeonGeers:
@fjansson There is no 'gold' standard, as far as I know. I did a minimum invasive action, so kg/ms^2 became kg/m/s^2, not so neat in itself either, but at least physically correct.

In TNO we use the format 'kg m^-1 s^-2' and I always assumed that came from CF conventions, but I never checked. I do know that LOTOS-EUROS uses the 'udunits' library (which does come from outside, I'll check the source), which requires a certain format. The unit parser I built uses the rules udunits requires, but I made it a lot more robust to also capture units that are not compliant to it.

Choosing a standard unit notation for DALES is a good idea and deserves attention, in my opinion.

PR #317, opened 2026-08-04
move some messages from used_profiles to modlogging:message()

fjansson:
also stop printing the nice orbital parameters calculated in RRTMG which currently aren't used anyway.


Xetalim (approved):
LGTM

